PR #14, opened 2020-09-02
Backends/cruxmir

fshaked:
Initial support for crux-mir.

This will break things with cargo-verify. In particular, when running cargo in compatibility-test you must specify the verifier like this: cargo test --features "verify propverify/verifier-crux"

### DIFF
--- a/compatibility-test/src/main.rs
+++ b/compatibility-test/src/main.rs
@@ -16,9 +16,9 @@
 
 #![allow(unused_imports)]
 
-#[cfg(not(verify))]
+#[cfg(not(feature = "verify"))]
 use proptest::prelude::*;
-#[cfg(verify)]
+#[cfg(feature = "verify")]
 use propverify::prelude::*;
 
 mod collections;
@@ -28,6 +28,7 @@ mod enumeration;
 
 // A simple test of the propverify/proptest library
 proptest! {
+    #[test]
     fn main(
         a in 4..8u32,
         b in 5..9u32)
--- a/propverify/src/strategy.rs
+++ b/propverify/src/strategy.rs
@@ -522,6 +522,7 @@ numeric_api! {
     isize;
 }
 
+#[cfg(feature = "float")]
 numeric_api! {
     f32;
     f64;
--- a/propverify/src/strategy.rs
+++ b/propverify/src/strategy.rs
@@ -153,28 +153,54 @@ pub fn prop_is_replay() -> bool {
     verifier::is_replay()
 }
 
+#[cfg(feature = "meta")]
 #[macro_export]
 macro_rules! proptest {
     (
-      $(#[$meta:meta])*
-      fn $test_name:ident($($parm:tt in $strategy:expr),+ $(,)?) $body:block
+        #[test]
+        $(#[$meta:meta])*
+        fn $test_name:ident($($parm:tt in $strategy:expr),+ $(,)?) $body:block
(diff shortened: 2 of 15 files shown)


verification-annotations/src/verifier_std.rs
```diff
@@ -0,0 +1,76 @@
...
+
+// Reject the current execution path with a verification success.
+// This is equivalent to assume(false)
+// and the opposite of report_error.
+//
+// Typical usage is in generating symbolic values when the value
+// does not meet some criteria.
+pub fn reject() -> ! {
+    // TODO: ???
+    panic!("reject")
```
alastairreid, line 45: I don't think that there is any reasonable way to implement this
so I worry that this will cause problems with strategies like 'char' that use this function.

verification-annotations/src/crux.rs
```diff
@@ -0,0 +1,93 @@
...
+// except according to those terms.
+
+/////////////////////////////////////////////////////////////////
+// FFI wrapper for Crux-mir static simulator tool
+/////////////////////////////////////////////////////////////////
+
+#[cfg(crux)]
+extern crate crucible;
+
+// pub mod crucible::symbolic;
```
alastairreid, line 16: Delete this commented out code?

propverify/src/strategy.rs
```diff
@@ -153,29 +153,55 @@ pub fn prop_is_replay() -> bool {
...
+        #[cfg_attr(crux,crux_test)]
+        // $(#[$meta])*
+        fn $test_name() {
+            $(
+                let str = stringify!($meta);
+                if str.starts_with("should_panic") {
+                    verifier::expect_raw(str);
+                }
+            )*
+            eprintln!("____META END____");
```
alastairreid, line 174: Can we change the string from "____META END____" to something that starts with "VERIFIER"?
Maybe "VERIFIER_META_END"?

verification-annotations/src/verifier_std.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright 2020 The Propverify authors
```
alastairreid, line 1: If I understand this correctly, this crate exists only to allow us to let us print the meta-annotations to support "should_panic".
In particular, I think it only exists to avoid linking errors?

So could we change every function body to just "panic()"?

compatibility-test/src/main.rs
```diff
@@ -16,9 +16,9 @@
 
 #![allow(unused_imports)]
 
-#[cfg(not(verify))]
+#[cfg(not(feature = "verify"))]
```
alastairreid, line 19: Is this change needed for Crux?

alastairreid (changes requested):
Can you
- drop the --cfg=verify -> --features="verify" change - unless it is needed by crux?
- change the ____META END___ string
- in verifier_std.rs, do one of
  - eliminate verifier_std.rs completely - it is not clear whether this is needed at all when in 'meta' mode
  - merge verifier_std.rs and verifier_crux.rs using conditional compilation
  - replace all function bodies with "panic()"
- squash changes that conflict with each other
   (It is hard to review a CL that makes a change in one commit and then changes it again in a later commit)

alastairreid:
Has all the code in this change been committed in other changes now?
If so, should we delete this pull request?

fshaked:
A better implementation was merged.

PR #16, opened 2020-09-05
fixing docs

chadbrewbaker:
There was a problem linking minisat pulled from apt-get (Ubuntu20.04), so I took it out of the package list.

Another of the make installs needed a sudo.

A make command was missing.


alastairreid (approved):
Thanks for these Chad - much appreciated!

PR #17, opened 2020-09-08
Add crux to verification-annotations

fshaked:
For example (after following the instructions in install-crux.md):

```shell
cd verification-annotations
RUSTFLAGS='--cfg test' cargo crux-test --features 'verifier-crux'
```

`assert!` reports the line number of `crux::verify` (so always the same line)!


fshaked:
Note that test t2 passes (should fail, multiplication overflow). This is a known issue of crux.

alastairreid (approved):
LGTM

PR #18, opened 2020-09-09
Add crux to propverify

fshaked:
For example:

```shell
cd compatibility-test
RUSTFLAGS='--cfg test --cfg verify' cargo crux-test --features 'verifier-crux' -- -t1
```

Some of the test fail (for the wrong reason), crux is pointing at strategy.rs:144:28, which is this call `Box::new(self)` in `Strategy::boxed`, with the message "unimplemented cast: Unsize".  In other case crux is pointing at ./lib/libcore/mem/maybe_uninit.rs:788:9 (crux's implementation of safe core) with the message "unimplemented cast: Misc".

### DIFF
--- a/compatibility-test/Cargo.toml
+++ b/compatibility-test/Cargo.toml
@@ -18,3 +18,4 @@ proptest = { version = "*" }
 
 [features]
 verifier-klee = ["propverify/verifier-klee"]
+verifier-crux = ["propverify/verifier-crux"]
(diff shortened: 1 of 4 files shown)


alastairreid (approved):
Unfortunate that unimplemented features in crux are sometimes seen as detecting an error.

Not sure what to do about it
- add extra tests that should not pass/fail?
- disable the tests that pass/fail for the wrong reason?

Let's ignore it for now and we can revisit later.

PR #20, opened 2020-09-09
Script fixes

alastairreid:


alastairreid:
I split this into two separate pull requests - combining them was silly

PR #22, opened 2020-09-10
Add assert macros to Klee

fshaked:
I implemented assert using eprintln and abort, with the same message string as I did for crux, assuming line number info will be added by Klee(?). Should I change that to just forward cond to std::assert?


alastairreid (approved):
Looks good.

(Not sure whether this is best done using eprintln/abort or by calling panic or by something else. But probably the only way to find out is to try this first and see where it takes us.)

PR #23, opened 2020-09-10
Clean the use of mod and use in tests

fshaked:
Note that `use crate as verifier` can be replaced with `use crate::*` (and remove `verifier::` everywhere).


alastairreid (approved):
Well, that's a lot easier!

PR #26, opened 2020-09-10
KLEE: treat linking errors as tool errors not SUT errors

alastairreid:
Tool errors are reported as UNKNOWN
while SUT errors are true errors reported as ERROR


fshaked:
LGTM

PR #42, opened 2020-10-07
Nondet trait

alastairreid:
Note: the file to read first is the last file: verification-annotations/src/traits.rs
This shows the new API that all the other changes are implementing/using

### DIFF
--- a/verification-annotations/src/crux.rs
+++ b/verification-annotations/src/crux.rs
@@ -10,16 +10,30 @@
 // FFI wrapper for Crux-mir static simulator tool
 /////////////////////////////////////////////////////////////////
 
-// Create an abstract value of type <T>
-//
-// This should only be used on types that occupy contiguous memory
-// and where all possible bit-patterns are legal.
-// e.g., u8/i8, ... u128/i128, f32/f64
-pub fn abstract_value<T: crucible::Symbolic>() -> T {
-    // We assume the string argument is just for reporting to the user, and
-    // doesn't affect the results.
-    T::symbolic("")
+use crate::traits::*;
+
+impl <T: AbstractValue> VerifierNonDet for T {
+    fn verifier_nondet(self) -> Self {
+        T::abstract_value()
+    }
 }
+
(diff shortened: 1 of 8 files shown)


verification-annotations/src/crux.rs
```diff
@@ -10,17 +10,31 @@
...
+    fn abstract_value() -> Self {
+        // We assume the string argument is just for reporting to the user, and
+        // doesn't affect the results.
+        let r : T = crucible::Symbolic::symbolic("");
+        r
+    }
 }
 
+impl <T: crucible::Symbolic + Default> Symbolic for T {
+    fn symbolic(_desc: &'static str) -> Self {
```
fshaked: Maybe pass str to symbolic below, instead of ""?

alastairreid: oops! Good catch.

verification-annotations/src/klee.rs
```diff
@@ -105,8 +147,8 @@ pub fn expect(msg: Option<&str>) {
 macro_rules! assert {
     ($cond:expr) => {
         if ! $cond {
-            eprintln!("VERIFIER: assertion failed: {}", stringify!($cond));
-            abort();
+            eprintln!("VERIFIER: panicked at 'assertion failed: {}', {}:{}:{}", stringify!($cond), std::file!(), std::line!(), std::column!());
```
fshaked, line 150: I'm not sure "panicked" fits here

alastairreid: This is a change to make verifier::assert work the same as std::assert for KLEE.
In particular, to make should_panic work, it needs to have "panicked at" in the message.

Another way to implement this change is to have this macro call 'crate::panic' and then 'crate::panic' would print the 'panicked at' part of the string

fshaked: Calling crate::panic makes it worse. I still think using "panicked" is not nice, but I guess nice is not our top priority.

fshaked (approved):

Issue #36, opened 2020-10-01
when use klee


StevenJiang1110:
Hello, I want to use klee to analyse some rust program and I set up my environment in a ubuntu18.04 docker following your docs. However, when I want to analyse the compiled file using `klee --output-dir=kleeout --warnings-only-to-file target/debug/deps/try_klee*.bc`, some error occured.The full error information is attached. I wonder how to fix it. Thanks a lot.
<img width="1784" alt="截屏2020-10-01 下午5 57 15" src="https://user-images.githubusercontent.com/27764680/94795278-b86b6a80-040f-11eb-946d-abef1b62d793.png">

StevenJiang1110:
when I install klee using `git merge nowack/handle_global_variables reid/lazy_intrinsic_rejection` in your doc, something seems incorrect. I wonder if this is about the error.
<img width="1252" alt="截屏2020-10-01 下午6 01 39" src="https://user-images.githubusercontent.com/27764680/94795793-7131a980-0410-11eb-9c63-3d17496eefc9.png">

StevenJiang1110:
My script to build docker image and to test is here. It should replay the error above.
[https://github.com/StevenJiang1110/work_env/tree/master/docker](url).

StevenJiang1110:
https://github.com/StevenJiang1110/work_env/tree/master/docker

StevenJiang1110:
I put my scripts and dockerfile on thie link, it should replay the error above. I hope I can get some idea to fix it. https://github.com/StevenJiang1110/work_env/tree/master/docker.

alastairreid:
Hi Steven,

Thanks for reporting these issues - we really appreciate it.

The biggest barrier to using these tools is that it is so hard to install them.
It would be great if you could send us a pull request for your dockerfile when you have something working.
I have barely used Docker myself so it would be great to get some help with this.

The stack dump is a direct result of the git merge failure.
The git merge failure was caused by the reid/lazy_intrinisic_rejection pull request being merged into the KLEE master branch.
I have fixed the instructions (waiting for review) to replace the merge instructions with

```
git remote add nowack https://github.com/MartinNowack/klee.git
git fetch --all
git merge nowack/handle_global_variables
```

I am told that the KLEE warnings about linking modules with different target tuples is probably benign.
I have opened a separate issue #37 for this.
(TL;DR: ideally, building rustc would also build a matching copy of clang. But it may not matter.)

StevenJiang1110:
[...]

PR #49, opened 2020-10-10
add dockerfile and scripts to build image for rust-verification-tools(klee only)

StevenJiang1110:
These scripts and dockerfiles is to build rust-verification-tools(klee only) docker image.
The building process almost follows the using-klee doc(three steps, install rust, install klee and install rust-verification-tools) but has some additional steps.(I think it may be essential)
First, the script compiles cargo from source to work as the default cargo for the compiled rustc, so it leads to a stage2 compiling. 
Second, the script installs rustup from official scripts and choose stage2 as the default toolchain.
The `try-klee` example has been added to the image for test use and it can work properly.  However, when I try to follow the example `replay input values` in doc , there is still one link error. I have mentioned this in #36 .


StevenJiang1110:
@googlebot I signed it!

alastairreid (commented):
Thanks a lot for this - we really appreciate it!

I like that it is split into a number of separate scripts so that it is easy to extend as we add support for new verification tools.

The main thing that needs to be changed is to change the dependency on the docker image "stevenjiang1110/ubuntu-expect".
If this is just ubuntu:18.04 plus "apt install expect", then this dependency is easy to remove.

[Added in edit] Another minor issue is that I need to use "sudo docker" on Linux but not on OSX. Is there a standard way of dealing with this? (Maybe I just need to use "sudo ./build" and that is enough.)

I am happy to make these changes myself or for you to make the changes - whichever you prefer.

docker/rust_verification/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM rust_klee:v3.0
```
alastairreid, line 1: Is this file necessary to be able to use the tools or is it just a demo?

I am new to Docker and do not know the usual ways of using Docker but I was hoping that we could package up the tools but still use them as if they were installed natively.
So I was expecting to see a Dockerfile that packaged up scripts/cargo-verify but I was not expecting one for try_klee

StevenJiang1110: This file is just a demo and is not necessary to be able to use the tools. What these dockerfiles and scripts try to build is more likely to be a development environment for rust-verification-tools, and the try-klee example is added to check whether  the environment has been installed properly and will not be exposed to user directly. This file can be deleted if we don't need the example to test environment.
Do you mean what you want is that when you have a native crate, for example ,hello , than you can run , `docker run rust-verification-tools  hello` to use scripts/cargo-verify in docker to verify it? It needs a small fix, and I can push a commit later for it.

StevenJiang1110:
As I know, the common practice is to add current user to docker group to avoid sudo privilege when using docker. run ` sudo groupadd docker`, `sudo usermod -aG docker $USER`, then exit current shell and reload. Docker then will no longer need sudo.

alastairreid:
Hi Steven,
Sorry it is taking me a while to review this. I haven't used Docker before (beyond trying 'hello world' like examples) so it is taking me a while to understand how Docker is meant to be used. Your PR is really, really helpful for building my understanding.

Thinking about what $USER to use, etc. my colleagues have pointed me at what our sister project developing the "Oak" runtime in Rust has done for Docker:

- When they build the Docker image, they pass in the USER, UID, GID that they want to use. [script](https://github.com/project-oak/oak/blob/main/scripts/docker_build)
- Within the Dockerfile, they use USER, UID, GID to create a user identical to the user who created the image. [Dockerfile](https://github.com/project-oak/oak/blob/5918a841ae9091bdc8ee2b5e72e5a7890bbd627b/Dockerfile#L254-L263)

The net effect of which is that if the image reads or writes a file, it does so using exactly the same user as the person that created the image.

I think this is the right way to go so I will push some additional changes that add this (for the rust_verification image). (Probably tomorrow and trying to copy their scripts as closely as possible.)

alastairreid:
I have amended the change as follows:

- remove the use of the external docker image stevenjiang1110/ubuntu-expect:lates
- remove the jjf user associated with that image

This is now fine to merge

PR #51, opened 2020-10-16
Amended docker support

alastairreid:
This is an amended version of https://github.com/project-oak/rust-verification-tools/pull/49
It removes external, unreviewable depenedencies and a hardcoded username 'jjf'


alastairreid:
Oops, this is the wrong way to modify someone else's PR during review

PR #61, opened 2020-10-29
Docker: seahorn support

alastairreid:
These dockerfiles build SeaHorn and its dependencies.

Currently untested - all I know is that installation does not fail.

### DIFF
--- a/docker/seahorn/Dockerfile
+++ b/docker/seahorn/Dockerfile
@@ -1,0 +1,66 @@
+FROM rvt_z3:latest
+
+USER root
+
+RUN echo ${Z3_DIR} && ls ${Z3_DIR}
+
+# Install Debian and Python dependencies
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get --yes update \
+  && apt-get install --no-install-recommends --yes \
+  build-essential \
+  clang-10 \
+  clang-format-10 \
+  clang-tools-10 \
+  cmake \
+  g++-7-multilib \
+  gcc-multilib \
+  git \
+  lib32stdc++-7-dev \
+  libboost-all-dev \
+  libgmp-dev \
+  libgmpxx4ldbl \
+  libncurses5-dev \
+  lld-10 \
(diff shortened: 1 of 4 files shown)


docker/seahorn/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+
+WORKDIR /opt
+
+ARG YICES_VERSION=2.6.2
+RUN curl --location https://yices.csl.sri.com/releases/${YICES_VERSION}/yices-${YICES_VERSION}-x86_64-pc-linux-gnu-static-gmp.tar.gz > yices.tgz \
+  && tar xf yices.tgz \
+  && rm yices.tgz
+
+ARG Z3_VERSION=4.8.7
+RUN curl --location https://github.com/Z3Prover/z3/releases/download/z3-${Z3_VERSION}/z3-${Z3_VERSION}-x64-ubuntu-16.04.zip > z3.zip \
```
fshaked: Two copies of z3? One here, and another one in the z3 dockerfile?

fshaked (commented):
Can you also add a checkout of https://github.com/yvizel/verify-c-common (those are scripts for running SH, don't need to build anything)?

docker/seahorn/Dockerfile
alastairreid: ok, deleting that

alastairreid: verify-c-common is already being cloned - just before seahorn.

The pattern is: do all the safe, unlikely to fail actions first (eg cloning repos) before starting on builds, etc..
This works a bit better with the docker cache when steps start failing.

alastairreid: ps For things like this, it is often simpler just to make the changes yourself and force-push them to this PR's branch.
In particular, if you need changes to how seahorn is built, you should either change this PR or accept this PR and create a PR on top of it - much faster than having me make the changes you need.

fshaked (approved):

PR #70, opened 2020-12-03
Bugfixes

alastairreid:
Two recent changes had not been tested with scripts/regression-test
They both introduced serious type errors.
The fix for one change is to add trait dependencies.
The fix for the other change is just to revert it - will need to rethink it.


fshaked (approved):
LGTM

PR #71, opened 2020-12-03
Docker: cleanup, use explicit versions, separate yices out

### LINKED ISSUE #68: docker/build fails
I'm getting some errors running the build script; the 'klee' container fails, complaining about a missing header:

```runtime/POSIX/CMakeFiles/RuntimePOSIX.dir/build.make:335: recipe for target 'runtime/POSIX/fd_init32_Release+Debug+Asserts.bc' failed
make[2]: *** [runtime/POSIX/fd_init32_Release+Debug+Asserts.bc] Error 1
In file included from /root/klee/klee/runtime/klee-libc/strncpy.c:37:                                                                                                                                             
/usr/include/stdlib.h:25:10: fatal error: 'bits/libc-header-start.h' file not found                                                                                                                               
#include <bits/libc-header-start.h>                                                                                                                                                                               
         ^~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                               
In file included from /root/klee/klee/runtime/klee-libc/putchar.c:10:                                                                                                                                             
/usr/include/stdio.h:27:10: fatal error: 'bits/libc-header-start.h' file not found                                                                                                                                
#include <bits/libc-header-start.h>                                                                                                                                                                               
         ^~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                               
In file included from /root/klee/klee/runtime/klee-libc/strcoll.c:10:                                                                                                                                             
/usr/include/string.h:26:10: fatal error: 'bits/libc-header-start.h' file not found                                                                                                                               
#include <bits/libc-header-start.h>                                                                                                                                                                               
         ^~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                               
In file included from /root/klee/klee/runtime/klee-libc/strtol.c:34:                                                                                                                                              
In file included from /usr/lib/llvm-10/lib/clang/10.0.0/include/limits.h:21:                                                                                                                                      
/usr/include/limits.h:26:10: fatal error: 'bits/libc-header-start.h' file not found                                                                                                                               
#include <bits/libc-header-start.h>                                                                                                                                                                               
         ^~~~~~~~~~~~~~~~~~~~~~~~~~                                                                                                                                                                               
1 error generated.                                                                                                                                                                                                
1 error generated.   
```

It looks like the scripts are just pulling from the master branch directly; do you have a SHA that's verified to work for klee?

alastairreid:
All git checkouts now use an explicit version number so that we get repeatable results as the projects we depend on are updated.

Fixes #68
(But you will also need the bugfixes in #70 because we had broken the regression tests.)

One thing that we considered doing as part of this change was
updating to the latest version of rustc (1.48.0).
We decided not to do this because it would require us to
update to LLVM-11 and the verification tools that we
use are possibly/probably not ready for LLVM-11 yet.

This change was tested by running:

    sudo docker system prune -a
    docker/build
    docker/run scripts/regression-test


fshaked (approved):
I committed a few changes.

PR #73, opened 2020-12-11
Add FFI support so that we can call C

### LINKED ISSUE #45: Crates that rely on C libraries are not supported
Crates that perform FFI calls to C libraries (i.e., libraries that are built separately from the cargo build system) are a problem because

1. The library is (probably) not being built as a bitcode file.

2. Even if we had bitcode files for those libraries, it is not clear how to link them with the Rust code.

The second part of the problem is the most tractable and would let us take a small number of really important libraries and futz with their build process to generate bitcode files. This would probably be quite painful so may not scale well.

The most general approach to the first problem might be to define a new target that treats LLVM bitcode files as the native representation. If that existed, than any reasonably portable package would "just work". But this is probably not the simplest approach (unless somebody has already done this).

See also #6.

alastairreid:
Adds the necessary magic to compile and link C code into the LLVM bitcode file that we use for verification.

The documentation describes the underlying commands, flags, etc. used to implement this.

Closes #45 

(Note that the base for this PR is the 'demos' branch from PR 72)


fshaked (approved):

alastairreid:
Aargh ... This PR was a change on top of another change. It looks as if submitting a PR on top of another change results in a merge into the other branch instead of a merge into the main branch.
I'll clean this up and avoid doing this again...

PR #76, opened 2021-01-15
RVT-patch-LLVM

alastairreid:
New tool for manipulating LLVM files as part of preparing them for verification


fshaked (approved):
Pushed a couple of small changes.

PR #77, opened 2021-01-18
ARGV: add support for std::env::args()

alastairreid:
This PR extends cargo-verify to be able to handle command line arguments, documents how it works and adds a simple example and regression test.


demos/simple/argv/src/main.rs
```diff
@@ -0,0 +1,25 @@
...
+#[cfg(verify)]
+use propverify::prelude::*;
+
+use std::env;
+
+/// Simple test to confirm that it is possible to pass arguments
+/// to Rust functions when verifying them.
+///
+/// Should pass:
+///    `cargo verify . foo`
```
fshaked: 'cargo run' passes all the arguments after '--' to the program. Can you make 'cargo verify' behave the same?

fshaked (approved):

PR #78, opened 2021-01-20
Use structopt instead of using clap directly

fshaked:
structopt can catch arg-parsing bugs at compile time.
structopt uses clap in the background.


alastairreid (approved):
That looks so much nicer than before!

PR #81, opened 2021-01-26
PropVerify: minor improvements

alastairreid:
Adds support for strategies of the form "x: i32" (which invokes Arbitrary).

Also modifies trait bounds to match proptest better.


fshaked (approved):
This doesn't allow mixing of the two styles. E.g. `fn test(x: i32, y in 0..100i32) ...` is allowed by proptest, but can't be parsed by this.

Good enough for now.

PR #82, opened 2021-01-29
Add support for SeaHorn.

fshaked:
Missing documentation, and integration with cargo-verify.


verification-annotations/lib/seahorn.c
```diff
@@ -0,0 +1,21 @@
+#include <stdlib.h>
+#include <stdint.h>
+
+void __VERIFIER_error() {}
```
alastairreid: I would have expected __VERIFIER_error to call a function to indicate that an error occurred?
(eg maybe "abort" or something?)

verification-annotations/src/seahorn.rs
```diff
@@ -166,6 +196,13 @@ macro_rules! assert_ne {
     }};
 }
 
+#[macro_export]
```
alastairreid: This is not provided by KLEE. Do you think we should add it?

scripts/cargo-verify-seahorn.sh
```diff
@@ -0,0 +1,178 @@
+#!/usr/bin/env bash
```
alastairreid, line 1: Is this script used instead of cargo-verify or is it called from cargo-verify?
(Or maybe something else?)

alastairreid, line 1: Can you add the copyright/license boilerplate to this file?

alastairreid (approved):
Approving this on basis that there is more to come.

fshaked:
Fixed everything mentioned above.

alastairreid (commented):
LGTM

PR #84, opened 2021-02-04
LLVM patcher: add --features to stub out feature tests

alastairreid:
Processor feature tests (std::std_detect::detect::*) return true if a processor feature (such as avx2) is present.
They are usually used to enable optimizations such as using vector instructions if the feature is present.

This code changes the feature detection functions to return false which has the effect of disabling the optimized path and, therefore, of avoiding processor dependent code paths.
And avoiding processor dependent code also tends to avoid embedded assembly and processor-specific intrinsic functions - which makes verification easier.


rvt-patch-llvm/src/main.rs
```diff
@@ -261,6 +283,70 @@ fn insert_call_at_head<'a>(
...
+// Function transformation functions
+////////////////////////////////////////////////////////////////
+
+/// Apply Rust mangling rule to a function name like 'foo::bar'.
+/// Does not insert a unique hash at the end or the final 'E' character.
+fn rust_mangle(n: &str) -> String {
+    // todo: I believe that this can be implemented by an elegant chain of
+    // iterators - but I am not smart enough to figure out the ownership, etc.
+    // issues that rustc reports so I went with the following code instead.
+    let mut rs = vec!["_ZN".to_string()];
```
fshaked: You can do it like this:
```rust
"_ZN".to_string() + 
&n.split("::").into_iter().map(|x| format!("{}{}", x.len(), x)).collect::<Vec<String>>().concat()
```
or like that:
```rust
n.split("::").into_iter().fold(vec!["_ZN".to_string()], |rs, x| { let mut rs = rs; rs.push(format!("{}{}", x.len(), x)); rs }).concat()
```

rvt-patch-llvm/src/main.rs
```diff
@@ -261,6 +283,70 @@ fn insert_call_at_head<'a>(
...
+    for x in n.split("::") {
+        rs.push(format!("{}{}", x.len(), x));
+    }
+    rs.concat()
+}
+
+/// Find a function whose name matches a Rust-mangled prefix
+fn find_mangled_function<'a>(module: &Module<'a>, prefix: &str) -> Option<FunctionValue<'a>> {
```
fshaked: There is somewhat similar function already, `get_function`, we should probably unify them.

rvt-patch-llvm/src/main.rs
```diff
@@ -261,6 +283,70 @@ fn insert_call_at_head<'a>(
...
+        if f.get_name().to_bytes().starts_with(prefix.as_bytes()) {
+            return Some(f);
+        }
+        of = f.get_next_function()
+    }
+    return None;
+}
+
+/// Delete the body of a function
+fn delete_body<'ctx>(f: &FunctionValue<'ctx>) {
```
fshaked: This is similar to what `replace_def_with_dec` does. 
Note that you must also call `f.remove_personality_function()` to keep llvm happy.

alastairreid: Good catch.
The goal of the new code is to replace the existing basic blocks with new code - so I don't want to remove personality functions, etc. in this case.

But reducing repeated code is good so I have refactored both functions as follows:

- Modified delete_body to use fun.get_basic_blocks based on your code in replace_def_with_dec
- Changed replace_def_with_dec to call delete_body.

rvt-patch-llvm/src/main.rs
alastairreid: I refactored these into get_function, get_function_by_regex and get_function_by_unmangled_name

rvt-patch-llvm/src/main.rs
alastairreid: Thanks. I went with the first version.
(I think I was previously trying to use &str - which lead to ownership issues.)

PR #86, opened 2021-02-05
FFI: add additional source of C-derived .bc files

alastairreid:
It turns out that .bc files can be in several different
places within target - so check all of them.

This change also generates the list of tests earlier when
using --tests so that we can report errors earlier.


fshaked (approved):

alastairreid:
I'm abandoning this change - I think it was a mistake

PR #88, opened 2021-02-09
Support for symbolic strings

alastairreid:
Composed of

- Efficient support for creating symbolic Vec<u8>.
- Tests involving strings
- Creating a new top-level directory 'runtime' to contain C support code required to interface between what the Rust library expects from the C library and the set of operations provided by different verifiers.
  - This will probably end up being a combination of code that is needed by all verifiers and code that is only needed with a subset of the verifier backends that we support. So it will end up requiring conditional compilation.
  - For now all it contains is support for allocating aligned memory blocks (used by crates like Regex).
  - In the near future, I will add code to support code that uses pthreads but only requires one thread. (i.e., to support almost every thread-safe library)
- Tests involving the standard Regex crate

### DIFF
--- a/verification-annotations/src/klee.rs
+++ b/verification-annotations/src/klee.rs
@@ -13,6 +13,8 @@
 pub use crate::traits::*;
 
 use std::default::Default;
+use std::ffi::CString;
+use std::mem;
 use std::os::raw;
 
 #[link(name = "kleeRuntest")]
@@ -118,6 +120,51 @@ make_verifier_nondet!(isize, get_value_isize);
 
 make_verifier_nondet!(f32, klee_get_value_f);
 make_verifier_nondet!(f64, klee_get_value_d);
+
+/// Allocate a symbolic vector of bytes
+pub fn verifier_nondet_bytes(n: usize) -> Vec<u8> {
+    // create an empty vector with capacity 'n'
+    let v: Vec<u8> = Vec::with_capacity(n);
+
+    // take vector apart so that we can access unused capacity
+    let mut v = mem::ManuallyDrop::new(v);
+    let p = v.as_mut_ptr();
+
+    unsafe {
+        // mark contents of vector as symbolic
+        let data = p as *mut raw::c_void;
+        let null = 0 as *const i8;
+        klee_make_symbolic(data, n, null);
+
+        // put vector back together again,
+        // with same capacity but increased length
+        Vec::from_raw_parts(p, n, n)
+    }
+}
+
+/// Allocate a symbolic CString
+pub fn verifier_nondet_cstring(size_excluding_null: usize) -> CString {
+    let mut r = verifier_nondet_bytes(size_excluding_null + 1);
+    for i in 0..size_excluding_null {
+        assume(r[i] != 0u8);
+    }
+    r[size_excluding_null] = 0u8;
+    unsafe { CString::from_vec_with_nul_unchecked(r) }
+}
+
+/// Allocate a symbolic ASCII String
+/// (ASCII strings avoid the complexity of UTF-8)
+pub fn verifier_nondet_ascii_string(n: usize) -> String {
+    let r = verifier_nondet_bytes(n);
+    for i in 0..n {
+        assume(r[i] != 0u8);
+        assume(r[i].is_ascii());
+    }
+    match String::from_utf8(r) {
+        Ok(r) => r,
+        Err(_) => reject(),
+    }
+}
 
 impl VerifierNonDet for bool {
     fn verifier_nondet(self) -> Self {
--- a/verification-annotations/src/tests.rs
+++ b/verification-annotations/src/tests.rs
@@ -136,6 +136,88 @@ fn concrete3() {
     verifier::assert!(b <= 1_000_000_000);
 }
 
+
+/// Test of verifier_nondet_bytes
+#[cfg_attr(not(feature = "verifier-crux"), test)]
+#[cfg_attr(feature = "verifier-crux", crux_test)]
+fn bytes() {
+    let a = verifier::verifier_nondet_bytes(8);
+    for i in &a {
+        verifier::assume(*i == 42);
+    }
+    if verifier::is_replay() {
+        println!("{:?}", a);
+    }
+    verifier::assert_eq!(a.len(), 8);
+    verifier::assert_ne!(a[2], 0u8);
+    verifier::assert_eq!(a[3], 42u8);
+}
+
+/// Test of verifier_nondet_cstring
+#[cfg_attr(not(feature = "verifier-crux"), test)]
+#[cfg_attr(feature = "verifier-crux", crux_test)]
+fn cstring() {
+    let a = verifier::verifier_nondet_cstring(8);
+
+    if verifier::is_replay() {
+        println!("{:?}", a);
+    }
+
+    // force string to be plain ASCII - to keep things simple
+    for i in a.as_bytes() {
+        // note: this code suffers from a path explosion and
+        // would benefit from using verifier::coherent!
+        // We will not do that though because the goal is to
+        // test this feature in isolation.
+        verifier::assume(i.is_ascii_alphabetic());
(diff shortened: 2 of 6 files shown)


alastairreid:
Note that this change should not be pushed for review until the script_fixes_2 change is merged.

verification-annotations/src/klee.rs
```diff
@@ -60,6 +62,51 @@ make_verifier_nondet!(isize);
...
+    }
+}
+
+/// Allocate a symbolic CString
+pub fn verifier_nondet_cstring(size_excluding_null: usize) -> CString {
+    let r = verifier_nondet_bytes(size_excluding_null + 1);
+    for i in 0..size_excluding_null {
+        assume(r[i] != 0u8);
+    }
+    assume(r[size_excluding_null] == 0u8);
```
fshaked: Wouldn't it be better to push the concrete value 0u8 instead of making it symbolic and assuming it is 0u8?

runtime/Makefile
```diff
@@ -0,0 +1,35 @@
...
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+# default target backend: override on command line
+TGT = klee
+
+# this rules comes first to make sure that it is the default target
```
fshaked: "rules" --> rule

runtime/Makefile
```diff
@@ -0,0 +1,35 @@
...
+default: librvt-$(TGT).a
+
+CC = clang-10
+CFLAGS = -flto=thin
+CFLAGS += -O1
+
+LINK = llvm-link-10
+
+C_SRCS = $(wildcard src/*.c)
```
fshaked: Use `:=` instead of `=` to make the RHS expression evaluate only once.
And the same for the line below.

verification-annotations/src/tests.rs
```diff
@@ -85,6 +86,110 @@ fn t5() {
...
+    if verifier::is_replay() {
+        println!("{:?}", a);
+    }
+
+    // force string to be plain ASCII - to keep things simple
+    for i in a.as_bytes().iter() {
+        verifier::assume(i.is_ascii_alphabetic());
+    }
+
+    for i in a.as_bytes().iter() {
```
fshaked: Do you need the `.iter()` here? `for` does that for you. You can also do all this in one assertion:
```
verifier::assert!(a.as_bytes().iter().all(char::is_ascii))
```

verification-annotations/src/klee.rs
alastairreid: Sure, I'll do that.

I think I was thinking about how to work with immutable objects - for which the assume is the right choice.
But it is easy to make it mutable in this case since I had only just created it.

runtime/Makefile
alastairreid: done

runtime/Makefile
alastairreid: There's always more to learn about Make!

Done

verification-annotations/src/tests.rs
alastairreid: I'm dropping the .iter() here and in the other place that I used '.as_bytes().iter()' a few lines above.

But I find that this point-free style less readable so I prefer the for loop.

verification-annotations/src/klee.rs
```diff
@@ -60,6 +62,51 @@ make_verifier_nondet!(isize);
...
+/// (ASCII strings avoid the complexity of UTF-8)
+pub fn verifier_nondet_ascii_string(n: usize) -> String {
+    let r = verifier_nondet_bytes(n);
+    for i in 0..n {
+        assume(r[i] != 0u8);
+        assume(r[i].is_ascii());
+    }
+    match String::from_utf8(r) {
+        Ok(r) => r,
+        Err(_) => reject(),
```
fshaked, line 165: Maybe this doesn't matter very much, but shouldn't this `reject()` be `unreachable()`?

verification-annotations/src/tests.rs
```diff
@@ -85,6 +86,110 @@ fn t5() {
     verifier::assert_ne!(a, a+1);
 }
 
+#[cfg_attr(not(feature = "verifier-crux"), test)]
+#[cfg_attr(feature = "verifier-crux", crux_test)]
+fn bytes() {
+    let a = verifier::verifier_nondet_bytes(8);
+    for i in a.iter() {
```
fshaked: `for i in &a {` will call `a.iter()`. Note the `&`, otherwise it calls `a.into_iter()` which moves `a` and you can no longer use it.

verification-annotations/src/tests.rs
```diff
@@ -85,6 +86,110 @@ fn t5() {
...
+#[cfg_attr(feature = "verifier-crux", crux_test)]
+fn string_ok() {
+    let a = verifier::verifier_nondet_ascii_string(6);
+
+    if verifier::is_replay() {
+        println!("{:?}", a);
+    }
+
+    // force string to be a legal int
+    for i in a.as_bytes().into_iter() {
```
fshaked: Don't need `.into_iter()`.

verification-annotations/src/tests.rs
```diff
@@ -85,6 +86,110 @@ fn t5() {
...
+fn string_should_fail() {
+    verifier::expect(Some("assertion failed"));
+    let a = verifier::verifier_nondet_ascii_string(6);
+
+    if verifier::is_replay() {
+        println!("{:?}", a);
+    }
+
+    // force string to be a legal int
+    for i in a.as_bytes().into_iter() {
```
fshaked: Don't need `.into_iter()`.

fshaked (changes requested):

verification-annotations/src/klee.rs
alastairreid: reject() is equivalent to assume(false) and so it prevents the verifier from reporting errors involving this path. (In KLEE, it silently halts the current path exploration.)

unreachable() generates an error that is treated as a verification error.

In this case, the function is meant to be generating a legal string so we would not want to report an error.

fshaked: But you already assume all the bytes are ascii, is the Err branch still reachable?

alastairreid: Ah, I see.

I am 99.9999% certain that this cannot fail.
But suppose I am wrong, then I would not want to report 'unreachable' because it would be an error in the verification support code not in the code being verified.

It would be good though to get some sort of report that there is a problem in the verification harness.
We should think about this a bit.

I think that this relates to a general problem in verification where you accidentally over-constrain the inputs so that you are not testing inputs that actually matter. In the extreme case (that is surprisingly common), the constraints are contradictory so there are no inputs and the verification is vacuously true.
I have been thinking that we should tackle that using a variation of what hardware engineers do when doing random testing: the System Verilog language lets you gather statistics about how many of the random inputs are in certain ranges - if a given range doesn't have enough values in it, you either need to test more or modify your test.
This would make sense as a proptest extension while for propverify we would just check that each range of interest is satisfiable.
(But exploring this is a problem for the future)

alastairreid:
Sorry this is taking a while to get back to - I wanted to get KLEE up and running with the new cargo-verify frontend before revisiting this.
It will probably be a few days before I get back to this.

alastairreid:
I acted on all comments apart from the one about using unreachable instead of reject.
(Neither is perfect (see discussion above) - but I think that reject is closer.)

Also, I moved the regex tests into demos/bottlenecks/regex since they were more about testing our ability to cope with language/tool features than they were about testing the verification annotations crate.

(Since I have fixed everything, I would normally self-approve. But that seems to be disabled.)

alastairreid:
(It wasn't disabled after all - confusing UI)

PR #89, opened 2021-02-15
kcachegrind support

alastairreid:
Consists of:

- rust2calltree program fixes name mangling in KLEE's kleeout/run.istats files
  (It should also be useful if you use any other coverage/profiling app with Rust code - but not tested that yet)
- Extend docker image with kcachegrind and prebuild rust2calltree

### DIFF
--- a/rust2calltree/src/main.rs
+++ b/rust2calltree/src/main.rs
@@ -1,0 +1,77 @@
+// Copyright 2020-2021 The Propverify authors
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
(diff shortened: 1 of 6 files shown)


rust2calltree/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+    // Open the output file
+    let output = opt.output.to_str().unwrap();
+    info!("Writing demangled output to {}", output);
+    let output = File::create(&output)
+        .with_context(|| format!("can't open output file '{}'", output))?;
+    let mut output = BufWriter::new(&output);
+
+    for line in input.lines() {
+        let mut line = line?;
+        if line.starts_with("fn=") {
```
fshaked: A bit nicer (don't have to change):
```rust
if let Some(s) = line.strip_prefix("fn=") {
  line = format!("fn={:#}", demangle(s));
} else ...
```
and the similar for the `else` part.

fshaked (approved):

rust2calltree/src/main.rs
alastairreid: Good idea - done.

PR #90, opened 2021-02-16
Rust cargo verify

fshaked:
Rewrite of cargo-verify in rust.
All the features were copied from the python script, but not very well tested.
Some of the command line arguments were changed, see --help.

### DIFF
--- a/cargo-verify/Cargo.toml
+++ b/cargo-verify/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "cargo-verify"
+version = "0.1.0"
+authors = ["Shaked Flur <sflur@google.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+cargo_metadata = "0.12.3"
+glob = "0.3.0"
+lazy_static = "1.4.0"
+log  = "0.4"
+num_cpus = "1.0"
+rayon = "1.5.0"
+regex = "1.4.3"
+rustc-demangle = "0.1"
+shell-escape = "0.1.5"
+stderrlog = "0.5"
+structopt = "0.3"
--- a/cargo-verify/src/main.rs
+++ b/cargo-verify/src/main.rs
@@ -1,0 +1,677 @@
+// Copyright 2020-2021 The Propverify authors
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+#![feature(command_access)]
+
+use std::{
+    collections::HashSet,
+    error, fmt,
+    path::{Path, PathBuf},
+    process::{exit, Command},
+    str::from_utf8,
+};
+
+use cargo_metadata::{CargoOpt, MetadataCommand};
+use glob::glob;
+use lazy_static::lazy_static;
+use log::error;
+use rayon::prelude::*;
+use regex::Regex;
+use rustc_demangle::demangle;
+use structopt::{clap::arg_enum, StructOpt};
+use utils::{add_pre_ext, Append};
+
+// utils must come before the other modules as it defines macros that they might
+// use.
+#[macro_use]
+mod utils;
+
+mod backends_common;
+mod klee;
+mod proptest;
+mod run_tools;
+mod seahorn;
+
+use run_tools::*;
+
+// Command line arguments
+#[derive(StructOpt)]
+#[structopt(
+    name = "cargo-verify",
+    about = "Execute verification tools",
+    // version number is taken automatically from Cargo.toml
+)]
+pub struct Opt {
+    // TODO: make this more like 'cargo test --manifest-path <PATH>'
+    // (i.e., path to Cargo.toml)
+    /// Path to Cargo.toml
+    #[structopt(
+        long = "manifest-path",
+        name = "PATH",
+        parse(from_os_str),
+        default_value = "Cargo.toml"
+    )]
+    cargo_toml: PathBuf,
+
+    /// Arguments to pass to program under test
+    #[structopt(name = "ARG", last = true)]
+    args: Vec<String>,
+
+    // backend_arg is used for holding the CL option. After parsing, if the user
+    // specified a backend it will be copied to the `backend` field below, if
+    // the user didn't specify a backend, we will auto-detect one, and put it
+    // in the `backend` field.
+    /// Select verification backend
+    #[structopt(
+        short = "b",
+        long = "backend",
+        name = "BACKEND",
+        possible_values = &Backend::variants(),
+        case_insensitive = true,
+    )]
+    backend_arg: Option<Backend>,
+
+    // See the comment of `backend_arg` above.
+    #[structopt(skip = Backend::Klee)] // the initial value has no meaning, it will be overwritten
+    backend: Backend,
+
+    /// Extra verification flags
+    #[structopt(long, number_of_values = 1, use_delimiter = true)]
+    backend_flags: Vec<String>,
+
+    /// Space or comma separated list of features to activate
+    #[structopt(long, name = "FEATURES", number_of_values = 1, use_delimiter = true)]
+    features: Vec<String>,
+
+    /// Run `cargo clean` first
+    #[structopt(short, long)]
+    clean: bool,
+
+    /// Verify all tests instead of 'main'
+    #[structopt(short, long)]
+    tests: bool,
+
+    // TODO: make this more like 'cargo test [TESTNAME]'
+    /// Only verify tests containing this string in their names
+    #[structopt(long, number_of_values = 1, name = "TESTNAME")]
+    test: Vec<String>,
+
+    // jobs_arg is used for holding the CL option. After parsing, if the user
+    // specified a value it will be copied to the `jobs` field below, if the
+    // user didn't specify a value, we will use num_cpus, and put it in the
+    // `jobs` field.
+    /// Number of parallel jobs, defaults to # of CPUs
+    #[structopt(short = "j", long = "jobs", name = "N")]
+    jobs_arg: Option<usize>,
+
+    // See the comment of `jobs_arg` above.
+    #[structopt(skip)]
+    jobs: usize,
+
+    /// Replay to display concrete input values
+    #[structopt(short, long, parse(from_occurrences))]
+    replay: usize,
+
+    /// Use verbose output (-vvvvvv very verbose output)
+    #[structopt(short, long, parse(from_occurrences))]
+    verbose: usize,
+}
+
+arg_enum! {
+    #[derive(Debug, PartialEq, Copy, Clone)]
+    enum Backend {
+        Proptest,
+        Klee,
+        Seahorn,
+    }
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum Status {
+    Unknown, // E.g. the verifier failed to execute.
+    Verified,
+    Error, // E.g. the verifier found a violation.
+    Timeout,
+    Overflow,
+    Reachable,
+}
+
+impl fmt::Display for Status {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Status::Unknown => write!(f, "UNKNOWN"),
+            Status::Verified => {
+                if f.alternate() {
+                    // "{:#}"
+                    write!(f, "OK")
+                } else {
+                    // "{}"
+                    write!(f, "VERIFIED")
+                }
+            }
+            Status::Error => write!(f, "ERROR"),
+            Status::Timeout => write!(f, "TIMEOUT"),
+            Status::Overflow => write!(f, "OVERFLOW"),
+            Status::Reachable => write!(f, "REACHABLE"),
+        }
+    }
+}
+
+type CVResult<T> = Result<T, Box<dyn error::Error>>;
+
+/// Parse the command line and make sure it makes sense.
+fn process_command_line() -> CVResult<Opt> {
+    // cargo-verify can be called directly, or by placing it on the `PATH` and
+    // calling it through `cargo` (i.e. `cargo verify ...`.
+    let mut args: Vec<_> = std::env::args().collect();
+    if args.get(1).map(AsRef::as_ref) == Some("verify") {
+        // Looks like the script was invoked by `cargo verify` - we have to
+        // remove the second argument.
+        args.remove(1);
+    }
+    let mut opt = Opt::from_iter(args.into_iter());
+    // let mut opt = Opt::from_args();
+
+    opt.backend = match opt.backend_arg {
+        // Check if the backend that was specified on the CL is installed.
+        Some(Backend::Proptest) => {
+            assert!(proptest::check_install());
+            Backend::Proptest
+        }
+        Some(Backend::Klee) => {
+            if !klee::check_install() {
+                Err("Klee is not installed")?;
+            }
+            Backend::Klee
+        }
+        Some(Backend::Seahorn) => {
+            if !seahorn::check_install() {
+                Err("Seahorn is not installed")?;
+            }
+            Backend::Seahorn
+        }
+        None => {
+            // If the user did not specify a backend, use the first one that we find.
+            let backend = if klee::check_install() {
+                Backend::Klee
+            } else if seahorn::check_install() {
+                Backend::Seahorn
+            } else {
+                assert!(proptest::check_install());
+                Backend::Proptest
+            };
+            println!("Using {} as backend", backend);
+            backend
+        }
+    };
+
+    // To be compatible with `cargo test`, features might be space separated.
+    opt.features = opt
+        .features
+        .iter()
+        .flat_map(|s| s.split(' '))
+        .map(String::from)
+        .collect::<Vec<_>>();
+
+    // Backend specific options.
+    match opt.backend {
+        Backend::Proptest => {
+            if opt.replay > 0 && !opt.args.is_empty() {
+                Err("The Proptest backend does not support '--replay' and passing arguments together.")?;
+            }
+        }
+        Backend::Seahorn => {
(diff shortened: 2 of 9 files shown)


cargo-verify/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "cargo-verify"
+version = "0.1.0"
+authors = ["Shaked Flur <sflur@google.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+structopt = "0.3"
```
alastairreid: Can you sort these alphabetically?
(Makes it more obvious where to add new dependencies and matches the 'use' lines)

cargo-verify/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+stderrlog = "0.5"
+cargo_metadata = "0.12.3"
+regex = "1.4.3"
+lazy_static = "1.4.0"
+rustc-demangle = "0.1"
+rayon = "1.5.0"
+num_cpus = "1.0"
+glob = "0.3.0"
+shell-escape = "0.1.5"
+# inkwell = { git = "https://github.com/alastairreid/inkwell", features = ["llvm10-0"] }
```
alastairreid: Remove this?

cargo-verify/src/utils.rs
```diff
@@ -0,0 +1,134 @@
...
+    ffi::{OsStr, OsString},
+    iter,
+    path::{Path, PathBuf},
+    process::Command,
+    str::Lines,
+};
+
+use log::info;
+
+// `info_at!(&opt, level, ...)` will print the formatted message `...` if
```
alastairreid: Should this and other comments use /// instead of // so that we could generate documentation?

cargo-verify/src/utils.rs
```diff
@@ -0,0 +1,134 @@
...
+// same Unicode scalar value is not available via Encoding in this crate."
+// The following is from https://stackoverflow.com/a/28175593
+pub fn from_latin1(s: &[u8]) -> String {
+    s.iter().map(|&c| c as char).collect()
+}
+
+// The Append trait lets you chain `append` calls where usually you would have
+// to mutate (e.g. using `push`).
+// Example:
+// assert_eq!(String::from("foo").append("bar"), { let mut x = String::from("foo"); x.push_str("bar"); x })
```
alastairreid: Should this be a doctest?

cargo-verify/src/utils.rs
```diff
@@ -0,0 +1,134 @@
...
+// Add `s` to the end of `self`, as a new component.
+impl Append<Path> for PathBuf {
+    fn append(mut self: PathBuf, s: impl AsRef<Path>) -> PathBuf {
+        self.push(s);
+        self
+    }
+}
+
+/// Example:
+/// assert_eq!(add_pre_ext(&PathBuf::from("foo.bar"), "baz"), PathBuf::from("foo.baz.bar"))
```
alastairreid, line 79: doctest?

cargo-verify/src/utils.rs
```diff
@@ -0,0 +1,134 @@
...
+
+// Add `s` to the end of `self`, as a new component.
+impl Append<Path> for PathBuf {
+    fn append(mut self: PathBuf, s: impl AsRef<Path>) -> PathBuf {
+        self.push(s);
+        self
+    }
+}
+
+/// Example:
```
alastairreid, line 78: Should this comment describe the function in words?

cargo-verify/src/proptest.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    utils::info_cmd(&cmd, "Proptest");
+
+    let output = cmd.output().expect("Failed to execute `cargo`");
+
+    let stdout = from_utf8(&output.stdout).expect("stdout is not in UTF-8");
+    let stderr = from_utf8(&output.stderr).expect("stderr is not in UTF-8");
+
+    if !output.status.success() {
```
alastairreid: Not critical to do now - but we should plan on doing a cleanup pass that factors out this error-handling then-branch so that it is shared with more functions.
Probably everything from line 41 (info_cmd) down to 57 (Ok(Status::Error)).

cargo-verify/src/seahorn.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// Return an int indicating importance of a line from KLEE's output
+// Low numbers are most important, high numbers least important
+//
+// -1: script error (always shown)
+// 1: brief description of error
+// 2: long details about an error
+// 3: warnings
+// 4: non-KLEE output
```
alastairreid: Change KLEE to SeaHorn?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    enum Backend {
+        Proptest,
+        Klee,
+        Seahorn,
+    }
+}
+
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub enum Status {
+    Unknown, // E.g. the varifier failed to execute.
```
alastairreid: verifier not varifier

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    Verified,
+    Error, // E.g. the varifier found a violation.
```
alastairreid: verifier not varifier

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    });
+
+    println!("VERIFICATION_RESULT: {}", status);
+    if status != Status::Verified {
+        exit(1);
+    }
+    Ok(())
+}
+
+// Compile a Rust crate to generate bitcode
```
alastairreid: Should this (and other comments in this file and other files) use /// instead of //?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+        let mains = mangle_functions(&opt, &bcfile, &[String::from(package) + "::main"])?;
+        match mains.as_slice() {
+            [(_, _)] => mains,
+            [] => Err("  FAILED: can't find the 'main' function")?,
+            _ => Err("  FAILED: found more than one 'main' function")?,
+        }
+    } else {
+        vec![("main".to_string(), "main".to_string())]
+    };
+    // Remove the package name from the function names (important for Klee?).
```
alastairreid: I couldn't figure out which part of cargo-verify this corresponds to

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    }
+
+    if opt.tests || !opt.test.is_empty() {
+        cmd.arg("--tests");
+    }
+
+    // The following line is not present because we care about the target It is
+    // there to allow us to use -Clto to build crates whose dependencies invoke
+    // proc_macros.
+    // FIXME: "=="?
```
alastairreid, line 519: This can only get fixed once cargo/rustc becomes less crazy.
Until then, we need this workaround.

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+
+    // Find the target directory
+    // (This may not be inside the crate if using workspaces)
+    info_at!(&opt, 4, "  Getting target directory");
+    let target_dir = get_meta_target_directory(&opt)?;
+
+    // {target_dir}/{target}/debug/deps/{package}*.bc
+    let bc_files = glob(
+        &glob::Pattern::escape(
+            target_dir
```
alastairreid, line 537: Would this chain of appends not be clearer as [target_dir, target, "debug", "deps", package]?
or as format!("{}/{}/debug/deps/{}*.bc", target_dir, target, package)?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+                    .map(|p| p.to_string_lossy())
+                    .collect::<Vec<_>>()
+                    .join(", ")
+            );
+            Err(format!("  FAILED: Test {} compilation error", &package))?
+        }
+    };
+
+    // {targetdir}/{target}/debug/build/ * /out/ *.o"
+    let c_files = glob(
```
alastairreid, line 581: see comment on previous glob

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    // print(json.dumps(build_plan, indent=4, sort_keys=True))
+    Ok((bc_file, c_files))
+}
+
+// Link multiple bitcode files together.
+fn link(out_file: &Path, in_files: &[PathBuf]) -> CVResult<()> {
+    let mut cmd = Command::new("llvm-link");
+    cmd.arg("-o").arg(out_file).args(in_files);
+
+    utils::info_cmd(&cmd, "llvm-link");
```
alastairreid: This pattern of error handling occurs a lot

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    )?
+    .filter_map(Result::ok)
+    .collect::<Vec<_>>();
+
+    // build_plan = read_build_plan(crate, flags)
+    // print(json.dumps(build_plan, indent=4, sort_keys=True))
+    Ok((bc_file, c_files))
+}
+
+// Link multiple bitcode files together.
```
alastairreid: I think this should be in common_backend.rs or utils.rs or similar.
All it does is call an existing utility program in a very generic way.

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    if !output.status.success() {
+        utils::info_lines("STDOUT: ", stdout.lines());
+        utils::info_lines("STDERR: ", stderr.lines());
+        Err("FAILED: Couldn't link")?
+    }
+
+    Ok(())
+}
+
+// Patch LLVM file to enable verification
```
alastairreid: Use /// instead of //?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+//
+// While this varies a bit according to the backend, some of the patching
+// performed includes
+//
+// - arranging for initializers to be executed
+//   (this makes std::env::args() work)
+// - redirecting panic! to invoke backend-specific intrinsic functions
+//   for reporting errors
+fn patch_llvm(opt: &Opt, options: &[&str], bcfile: &Path, new_bcfile: &Path) -> CVResult<()> {
```
alastairreid, line 608: Move to utils.rs or similar?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    utils::info_lines("STDERR: ", stderr.lines());
+
+    if !output.status.success() {
+        Err("FAILED: Couldn't run rvt-patch-llvm")?
+    }
+
+    Ok(())
+}
+
+fn get_meta_package_name(opt: &Opt) -> CVResult<String> {
```
alastairreid: Add /// comment.

Move to utils.rs?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+                // Anything else will be replaced with the '_' character.
+                _ => true,
+            },
+            "_",
+        );
+
+    Ok(name)
+}
+
+fn get_meta_target_directory(opt: &Opt) -> CVResult<PathBuf> {
```
alastairreid: Add ///
Move to utils.rs?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+    let dir = MetadataCommand::new()
+        .manifest_path(&opt.cargo_toml)
+        .features(CargoOpt::SomeFeatures(opt.features.clone()))
+        .exec()?
+        .target_directory;
+
+    Ok(dir)
+}
+
+fn get_default_host(crate_path: &Path) -> CVResult<String> {
```
alastairreid: Add ///
Move to utils.rs?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+        .map(|l| l.split(" ").collect::<Vec<_>>())
+        .filter(|l| l.len() == 3 && l[1] == "T" && fs.iter().any(|f| f == &l[2]))
+        .count();
+
+    info_at!(&opt, 4, "    Found {} functions", count);
+    count
+}
+
+// Generate a list of tests in the crate by parsing the output of `cargo test --
+// --list`
```
alastairreid: Unfortunate line break in middle of command.
Consider splitting line earlier?

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,841 @@
...
+        .filter_map(|l| {
+            TEST.captures(l)
+                .map(|caps| caps.get(1).unwrap().as_str().into())
+        })
+        .collect();
+
+    Ok(tests)
+}
+
+// Find a function defined in LLVM bitcode file
```
alastairreid: The comment for the function with the same name in Python script talks about what happens with the hash at the end of the mangled name.

More broadly, I think it would be best to copy the comments from the Python script exactly as they are and then improve them instead of deleting them or writing something else.

alastairreid (commented):
A few general things that I commented on a few times but that apply in many places in the code.

- use /// comments before functions and add a comment before every function
  - if the Python code had a comment before a function, copy that over and improve/update it as appropriate.
- There are a bunch of functions that just execute a program and report errors. I think these should be in utils.rs or backends_common.rs unless they do something interesting.
- A lot of code that executes an external program has a similar pattern of code to report the error - roughly starting from "info_cmd" and continuing down until the error has been reported. It would be good to factor that out into a macro or helper function.

fshaked:
>   * if the Python code had a comment before a function, copy that over and improve/update it as appropriate.

I mostly did that, I might have missed one or two, I will check.

> * There are a bunch of functions that just execute a program and report errors. I think these should be in utils.rs or backends_common.rs unless they do something interesting.

I disagree, those are cargo-verify's core functionality; I think utils.rs is for things we wish were in std.

> * A lot of code that executes an external program has a similar pattern of code to report the error - roughly starting from "info_cmd" and continuing down until the error has been reported. It would be good to factor that out into a macro or helper function.

Ok.

alastairreid:
> > * There are a bunch of functions that just execute a program and report errors. I think these should be in utils.rs or backends_common.rs unless they do something interesting.
> 
> I disagree, those are cargo-verify's core functionality; I think utils.rs is for things we wish were in std.

I don't think they should be in main.rs.  I don't care if that is one of the two files I mentioned or a new file.
But I think they are cluttering up main.rs.

fshaked:
> > > * There are a bunch of functions that just execute a program and report errors. I think these should be in utils.rs or backends_common.rs unless they do something interesting.
> > 
> > 
> > I disagree, those are cargo-verify's core functionality; I think utils.rs is for things we wish were in std.
> 
> I don't think they should be in main.rs. I don't care if that is one of the two files I mentioned or a new file.
> But I think they are cluttering up main.rs.

Ok

cargo-verify/src/utils.rs
fshaked: Probably, but we will have to wrap the code in a library.

cargo-verify/src/proptest.rs
fshaked: Done.

cargo-verify/src/main.rs
fshaked: The line below the comment :)
The problem starts with the name "tests" in the code, what it should really be is "functions", because in general we check functions (at the moment it can be `main` or one of the test functions).
Now, what the comment is trying to say is that after we get the mapping (`tests` maps function names to mangled names), we remove from each key the prefix `<package-name>::`, if the key has such prefix, because, I think, when we pass this name to Klee it expects it to be without the prefix.

cargo-verify/src/main.rs
fshaked: I don't understand what is going on here. We are passing `--target=={target}` to `cargo build`? What is the effect of this?
I changed this `==` to `=`, should I change it back?

cargo-verify/src/main.rs
fshaked: target_dir is not printable (need to use `target_dir.to_string_lossy()`); I also need to pass everything to `escape`, incase they use something that might be interpreted by glob as something else, but I must not pass the `*` through escape. So I don't think there is something significantly better that I can do here.

Initially I did this without using the globe crate and it was amazingly ugly.

cargo-verify/src/main.rs
fshaked: It is now just one line (actually cargo fmt broke it over multiple lines) in it's inlined (no function).

cargo-verify/src/main.rs
fshaked: Sorry about the comments. I mostly tried to do just that, at least in my head I thought I was doing that :).

fshaked:
I fixed almost everything mentioned above. I didn't fix the doctests because it needs a workaround to work with bin crates, and I didn't fix the complex use of glob because I don't think it's possible (see replies above).

I also incorporated most of the small commits @alastairreid did to the rcv branch, some by using `git cherry-pick` and some by hand. I didn't change Klee's output dir back to `kleeout-{test}` (2ea8598) because that will litter the crate dir if you have a lot of tests. And instead of renaming the files we generate (f602cf6) I filter the .bc files looking for exactly one '.'.

cargo-verify/src/main.rs
alastairreid: I think my comment was a response to an older? comment in the code about the whole idea of using --target, (not = vs ==).

= is the right thing to use - == must be a typo by me that cargo fails to detect as an error

cargo-verify/src/main.rs
```diff
@@ -0,0 +1,677 @@
...
+                .append("debug")
+                .append("deps")
+                .append(package)
+                .to_str()
+                .ok_or("not UTF-8")?,
+        )
+        .append("*.bc"),
+    )?
+    .filter_map(Result::ok)
+    // Filter only files that have exactly one '.'
```
alastairreid, line 549: Can you make this comment better?
I think you are doing this to distinguish between files created by Cargo and files created by us (from patch or link) - but I worry that we will forget that really important detail!

alastairreid (commented):
If you can improve that one comment about the really subtle filenaming choice/filtering,  I'll approve this.

alastairreid (approved):
LGTM

PR #91, opened 2021-02-19
Rcv docker

alastairreid:
Mostly docker-related changes to switch over to the new cargo-verify version


docker/build
```diff
@@ -11,6 +11,11 @@ set -e
 rm -rf docker/rvt/rvt-patch-llvm
 cp -r `pwd`/rvt-patch-llvm docker/rvt/rvt-patch-llvm
 
+# Copy cargo-verify over
+(cd cargo-verify; cargo clean)
```
fshaked: Alternatively (you don't have to change):
```shell
cargo clean --manifest-path cargo-verify/Cargo.toml
```

docker/build
```diff
@@ -11,6 +11,11 @@ set -e
...
+rm -rf docker/rvt/cargo-verify
+cp -r `pwd`/cargo-verify docker/rvt/cargo-verify
```
fshaked: (don't have to change) `cp -a` is more appropriate.

docker/rvt/Dockerfile
```diff
@@ -51,5 +57,9 @@ RUN rustup toolchain link stage1 ${RUSTC_DIR}/build/x86*/stage1 \
 COPY --chown=${USER_UID}:${USER_GID} ./rvt-patch-llvm ${USER_HOME}/rvt-patch-llvm
 RUN cargo install --root=${USER_HOME} --path=${USER_HOME}/rvt-patch-llvm
 
+# Build cargo-verify too
+COPY --chown=${USER_UID}:${USER_GID} ./cargo-verify ${USER_HOME}/cargo-verify
+RUN rustup run nightly cargo install --root=${USER_HOME} --path=${USER_HOME}/cargo-verify
```
fshaked: `cargo +nightly install ...` is cleaner?

alastairreid: Cleaner but doesn't actually work

```
$ cargo +nightly install  --root=${USER_HOME} --path=${USER_HOME}/cargo-verify
error: no such subcommand: `+nightly`
```

docker/build
alastairreid: Done

docker/build
alastairreid: Done

docker/rvt/Dockerfile
fshaked: Because of this [line](https://github.com/project-oak/rust-verification-tools/blob/3147f403697f89fc85d94522a9d36965adb19dc5/docker/rvt/Dockerfile#L32) we are using the cargo we built from source with rust, and it doesn't have this feature. Change that line to `ENV PATH="${USER_HOME}/.cargo/bin:$PATH"`, and `cargo +nightly ...` should work.

alastairreid: I think that we should make the Rust that works with verification be the default Rust (i.e., the one on the path).
So I do not want to change the PATH.

(What this means for us is that when compiling cargo-verify, we have one slightly awkward command to compile with instead of a different slightly awkward command. Not much to choose between them in my opinion.)

fshaked: We indead should make the Rust that work with verification the default Rust, but we should let rustup manage that for us. So don't put rust in the PATH directly, only put ~/.cargo/bin in PATH (which is what rustup requires) and `rustup default stage1` will make the rust we built for verification the default.

alastairreid: Ah, I see what you mean.
I've turned the PATH round so that ~/.cargo/bin is nearer start of path - which lets us use cargo +nightly

fshaked (approved):
LGTM

PR #93, opened 2021-02-19
Runtime library

alastairreid:
This library fills gaps in the C-library support provided by verifiers and required by Rust libraries.

To support this, we now invoke llvm-link every time we build, not just if the program under test uses the FFI to call C code.


runtime/Makefile
```diff
@@ -0,0 +1,35 @@
...
+OBJS := $(patsubst src/%.c, build_$(TGT)/%.o, $(C_SRCS))
+
+build_$(TGT)/%.o: src/%.c
+	mkdir -p build_$(TGT)
+	$(CC) $(CFLAGS) $^ -c -o $@
+
+rvt-$(TGT).bc: $(OBJS)
+	$(LINK) $(OBJS) -o $@
+
+clean::
```
fshaked, line 31: Double colon?

cargo-verify/src/main.rs
```diff
@@ -404,31 +404,36 @@ fn verifier_run(opt: &Opt, bcfile: &Path, name: &str, entry: &str) -> Status {
...
-            .arg(&new_bc_file)
-            .arg(&bc_file)
-            .args(&c_files)
-            .output_info()?;
-        bc_file = new_bc_file;
-    }
+    // Link bc file (from all the Rust code) against the (backend-specific)
+    // runtime library and any c_files generated by build scripts from any C/C++
+    // build scripts
+    let new_bc_file = bc_file.with_file_name("link.bc");
```
fshaked: To be consistent change it back to:
```rust
let new_bc_file = add_pre_ext(&bc_file, "link");
```

runtime/Makefile
alastairreid: Allows you to have multiple rules with the same name.
(Especially important in longer Makefiles)

cargo-verify/src/main.rs
alastairreid: oops - careless rebase.

fshaked:
LGTM

PR #94, opened 2021-02-19
Cargo-verify: always patch with --features

alastairreid:
This makes processor-specific feature tests return false so that,
for example, vectorized versions of code are disabled.

This is probably good to use with SeaHorn as well - but I am
only using this with KLEE because I have not tested with SeaHorn.

### DIFF
--- a/cargo-verify/src/main.rs
+++ b/cargo-verify/src/main.rs
@@ -437,8 +437,10 @@ fn build(opt: &Opt, package: &str, target: &str) -> CVResult<PathBuf> {
         bc_file = new_bc_file;
     }
 
-    if !opt.args.is_empty() {
-        info_at!(&opt, 1, "  Patching LLVM file for initializers");
+    // todo: This is probably useful with all verifiers - but
+    // making it KLEE-only until we have a chance to test it.
+    if opt.backend == Backend::Klee {
+        info_at!(&opt, 1, "  Patching LLVM file for initializers and feature tests");
         let new_bc_file = add_pre_ext(&bc_file, "patch-init-feat");
         patch_llvm(
             &opt,


cargo-verify/src/main.rs
```diff
@@ -437,8 +437,10 @@ fn build(opt: &Opt, package: &str, target: &str) -> CVResult<PathBuf> {
         bc_file = new_bc_file;
     }
 
-    if !opt.args.is_empty() {
-        info_at!(&opt, 1, "  Patching LLVM file for initializers");
+    // todo: This is probably useful with all verifiers - but
+    // making it KLEE-only until we have a chance to test it.
+    if opt.backend == Backend::Klee {
```
fshaked, line 442: There's already a similar check in `process_command_line()`.

alastairreid: You mean this code:
```
            if !opt.args.is_empty() {
                Err("The Seahorn backend does not support passing arguments yet.")?;
            }
```
???

This PR fixes problems with crates that have processor-feature-dependent optimizations (eg hand vectorized code). We want to do this whether the verification uses command line arguments or not.
I strongly suspect that we will want to do this with SeaHorn too and it will just work - but I didn't want to risk breaking things that I had not tested.


While making this change, I also made it apply the initializer transformation to code even if you are not using command line arguments.
I should probably have done this from the beginning - if you have any other initializers in your codebase it would be kind of surprising (i.e., annoying to debug) that they only got invoked if you pass arguments to the program.
(This is the change that may make SeaHorn less happy - I have no idea whether the initializers are likely to cause problems for SeaHorn.)

(Unless I hear otherwise, I will not change this commit.)

fshaked:
LGTM

PR #95, opened 2021-02-22
New cargo-verify fixes

alastairreid:
Two minor fixes:

- Don't optimize (even if using KLEE). It slows the build process and doesn't seem to help verification step. 
- 'cargo test -- --list' started failing complaining about not being able to build doctests 
  because we did not install the rustdoc component for our stage1 compiler.
  Prevent this error by not trying to verify doctests.


alastairreid:
oops, one of these commits is broken - should have tested better

PR #96, opened 2021-02-23
KLEE-related fixes

alastairreid:
These changes get scripts/regression-test to pass.
(This script tests KLEE - next step will be extending the script to also check SeaHorn.)

- Don't optimize even when using KLEE (turns out it was not always beneficial)
- Prevent 'cargo verify --tests' from complaining that rustdoc is not installed.
  (Installing rustdoc does not solve this problem, getting 'cargo test -- --tests' not to consider doc tests is the fix)
- Fix minor problem involving creation of kleeout directory
- Update gitignore rules to handle kleeout directory
- In docker, ensure that dumping core files is off by default.
   (I don't know when I ever found core files useful)
- Add 'docker/init' script that rebuilds all the tools/libraries that we depend on.
   (This doesn't get used while building docker images - I haven't figured that out yet. But it is really useful to use within a running docker container)

### DIFF
--- a/cargo-verify/src/klee.rs
+++ b/cargo-verify/src/klee.rs
@@ -26,8 +26,11 @@ pub fn check_install() -> bool {
 
 /// Run Klee and replay
 pub fn verify(opt: &Opt, name: &str, entry: &str, bcfile: &Path) -> CVResult<Status> {
-    let out_dir = opt.cargo_toml.with_file_name("kleeout").append(name);
+    // KLEE output files are put in kleeout directory with filename `name`
+    let klee_dir = opt.cargo_toml.with_file_name("kleeout");
+    fs::create_dir_all(&klee_dir)?;
 
+    let out_dir = klee_dir.append(name);
     // Ignoring result. We don't care if it fails because the path doesn't
     // exist.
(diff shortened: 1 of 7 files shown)


cargo-verify/src/klee.rs
```diff
@@ -26,8 +26,11 @@ pub fn check_install() -> bool {
 
 /// Run Klee and replay
 pub fn verify(opt: &Opt, name: &str, entry: &str, bcfile: &Path) -> CVResult<Status> {
-    let out_dir = opt.cargo_toml.with_file_name("kleeout").append(name);
+    // KLEE output files are put in kleeout directory with filename `name`
+    let klee_dir = opt.cargo_toml.with_file_name("kleeout");
+    fs::create_dir_all(klee_dir.clone())?;
```
fshaked: Unnecessary clone, pass `&klee_dir` instead.

fshaked:
>   (This doesn't get used while building docker images - I haven't figured that out yet. But it is really useful to use within a running docker container)

You probably want to call it from `docker/build`, not from any of the Dockerfiles.

fshaked (approved):

PR #99, opened 2021-02-24
fix(seahorn): call assume with right args

priyasiddharth:


priyasiddharth:
@googlebot I signed it!

priyasiddharth:
@fshaked can you PTAL? Thanks

fshaked:
Thanks for submitting.

Most of the changes are not relevant, and will be fixed by another pull request @alastairreid is working on.

Do we still need the `if` around the call to `__VERIFIER_assume`? If we keep the `if` does the fix really change anything?

verification-annotations/src/seahorn.rs
```diff
@@ -38,7 +38,7 @@ pub fn abort() -> ! {
 /// Any paths found must satisfy this assumption.
 pub fn assume(pred: bool) {
     if ! pred {
-        unsafe { __VERIFIER_assume(0); }
+        unsafe { __VERIFIER_assume(pred as i32); }
```
alastairreid, line 41: Is there an efficiency difference in SeaHorn between

"if !pred { unsafe { __VERIFIER_assume(0)} }"

and 

"__VERIFIER_assume(pred as i32)"

I think they mean the same but the latter maybe creates less branches/merges?

cargo-verify/src/main.rs
```diff
@@ -429,7 +429,7 @@ fn build(opt: &Opt, package: &str, target: &str) -> CVResult<PathBuf> {
     Command::new("llvm-link")
         .arg("-o")
         .arg(&new_bc_file)
-        .arg(runtime)
+        //.arg(runtime)
```
alastairreid, line 432: I have just merged a fix for this problem into the main branch.
After starting docker, run 'docker/init' which will build the runtime for both KLEE and SeaHorn as well as building the cargo-verify and the rvt-patch-llvm tools.

alastairreid (commented):
Thanks for sharing this with us - I think we have now fixed these issues.

alastairreid:
I think we have fixed all these - please reopen if we missed something.

verification-annotations/src/seahorn.rs
priyasiddharth: Sorry I missed reading the `if`. This is not needed. 

thanks

PR #100, opened 2021-02-24
(Seahorn) Handle panic earlier

fshaked:
This allows Seahorn to handle `std::assert!`, and other cases of panic as `abort`.

### DIFF
--- a/rvt-patch-llvm/src/main.rs
+++ b/rvt-patch-llvm/src/main.rs
@@ -56,10 +56,6 @@ struct Opt {
     #[structopt(short, long, conflicts_with = "initializers")]
     seahorn: bool,
 
-    /// Select a specific test to run (instead of 'main')
-    #[structopt(short, long, name = "TEST")]
-    test: Option<String>,
-
     /// Increase message verbosity
     #[structopt(short, long, parse(from_occurrences))]
     verbosity: usize,
@@ -106,9 +102,9 @@ fn main() {
     }
 
     if opt.seahorn {
-        handle_main(&module, &opt.test);
+        handle_main(&module);
 
-        handle_panic(&module);
+        handle_panic(&context, &module);
 
         replace_def_with_dec(
             &module,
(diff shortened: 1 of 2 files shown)


rvt-patch-llvm/src/main.rs
```diff
@@ -392,12 +392,36 @@ where
...
     // TODO: make "spanic" a CL arg.
     if let Some(spanic) = module.get_function("spanic") {
-        if let Some(unwind) = module.get_function("rust_begin_unwind") {
-            unwind.replace_all_uses_with(spanic);
-            info!("Replaced panic handling ('rust_begin_unwind') with 'spanic'.");
+        // Collect all the functions that start with
+        // "_ZN3std9panicking11begin_panic17h" (there will be more than one).
+        let mut panics = vec![];
+        let mut op_fun = module.get_first_function();
+        while let Some(fun) = op_fun {
```
alastairreid: I wondered if this could be implemented using get_function_by_unmangled_name("std::panicking::begin_panic").

I think the answer is 'no' if there can be more than one match.
But maybe what we really want here is a variant of get_function_by_* that returns a Vec or an Iterator?

alastairreid: Also... I don't understand why there is more than one - what is the difference between them?
And how can there be more than one - do the different versions found differ only in the hash at the end?

rvt-patch-llvm/src/main.rs
```diff
@@ -392,12 +392,36 @@ where
...
+                .unwrap_or(false)
+            {
+                panics.push(fun);
+            }
+            op_fun = fun.get_next_function();
+        }
+
+        // Delete the body of each panic function and replace it with a call to
+        // `spanic`.
+        let builder = context.create_builder();
```
alastairreid, line 380: I think that this is almost exactly what stub_functions does.
Specifically, I would think of writing it like this:
```
stub_functions(&context, &module, &functions, |builder| { builder.build_call(spanic, &[], "call"); builder.build_return(None); })
```

The only problem is that the second argument (&functions) takes a list of function names as an argument.

Maybe stub_functions should be changed so that the third argument (&functions) is replaced with either
- a vector/iterator of FunctionValue
or
- a predicate to test function names "Fn(String) -> bool"

alastairreid (approved):
I've added some suggestions for how this could be refactored to reuse existing parts of rvt-patch-llvm.

But I'm approving this so that you can decide whether to do that refactoring now or  delay it to later.

rvt-patch-llvm/src/main.rs
fshaked: I changed the get_function_* functions to return Vec instead of Option.

> Also... I don't understand why there is more than one - what is the difference between them?

I don't understand either. I see multiple instances with different hash suffix. They all have the same body.

fshaked:
@alastairreid, I pushed a big fix, please review again.

rvt-patch-llvm/src/main.rs
```diff
@@ -337,67 +331,73 @@ where
 // Transformations associated with SeaHorn
 ////////////////////////////////////////////////////////////////
 
-fn handle_main(module: &Module, otest: &Option<String>) {
-    // Remove the main function rustc generates.
+fn handle_main(module: &Module) {
+    // Remove the main function rustc generates. I don't know why, but passing
+    // --entry=.. to seahorn is not enough, we also have to remove main.
```
alastairreid, line 336: :-)

alastairreid (approved):

PR #104, opened 2021-03-04
Add `--script` for generating a script

fshaked:
Add two new fields to `Opt`, `script_arg` to hold the user option (value from CL), and `script` which is a `Mutex` protecting a `File` so we can write to the script from concurrent jobs.

The new function `add_to_script` does the actual writing to the script, everytime cargo-verify executes a command.

### DIFF
--- a/cargo-verify/src/main.rs
+++ b/cargo-verify/src/main.rs
@@ -11,9 +11,11 @@
 use std::{
     collections::HashSet,
     error, fmt,
+    fs::{self, File, OpenOptions},
     path::{Path, PathBuf},
     process::{exit, Command},
     str::from_utf8,
+    sync::Mutex,
     time::Instant,
 };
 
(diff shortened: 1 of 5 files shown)


cargo-verify/src/main.rs
```diff
@@ -120,6 +122,18 @@ pub struct Opt {
     /// Use verbose output (-vvvvvv very verbose output)
     #[structopt(short, long, parse(from_occurrences))]
     verbose: usize,
+
+    // script_arg is used for holding the CL option. After parsing, if the user
+    // specified a script, a `File` will be opened for writing, wrapped in a
+    // `Mutex` to allow concurrent jobs to write to it, and put in the `script`
+    // field below.
+    /// Generate a script with all the commands (and environment variables) that cargo-verify runs
+    #[structopt(long = "script", name = "NAME")]
```
alastairreid: Would PATH or FILENAME be a better label than NAME?
(I see PATH and FILENAME being used in "rustc --help" for example)

alastairreid (approved):
Approving this - but you might want to tweak the option name (see comment)

cargo-verify/src/main.rs
fshaked: Changed to PATH.

PR #106, opened 2021-03-05
The `name` method in structopt is not what you think it is

fshaked:
`name` is similar to `Arg::with_name` in clap, which has to be unique.


alastairreid:
Intrigued, I tried to find out what the difference was and... the #[structopt] documentation doesn't mention value_name at all. So I'm not sure what it is.

alastairreid (approved):

fshaked:
structopt passes everything (except for a few magic methods, like name) to Clap.
You can find value_name [here](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.value_name).

PR #107, opened 2021-03-08
KLEE extensions for path merging and concretization

alastairreid:
Adds the following functions

- `concretize!` to merge paths within a block
- `is_symbolic(x)` to test if a value x is symbolic or concrete
- `get_concrete_value(x)` to solve the current constraints on a value x and return a concrete value (this is most useful on symbolic values :-))
- `concretize(x)` to exhaustively solve the current constraints on a value x
  - forks separate paths for each resulting concrete value
  - since concretize is exhaustive, this may result in far more paths than you want
- `sample(N, x)` to solve the current constraints on a value up to N times
  - forks up to N separate paths for each resulting concrete value
  - since sample is not exhaustive, sample does not explore all possible behaviours of the program

At the moment, these extensions are only supported by the KLEE backend.
It may not be possible to implement these extensions for all backends.

### DIFF
--- a/demos/bottlenecks/merging/src/main.rs
+++ b/demos/bottlenecks/merging/src/main.rs
@@ -1,0 +1,85 @@
+/// Test the impact of path merging.
+///
+/// This is a scaling test to explore path explosions
+/// using the canonical symbolic execution example of
+/// a loop with multiple branches in it which results
+/// in 2^N paths unless merging is used.
+///
+/// Running this test on KLEE with
+///
+///     cargo verify --backend=klee --tests -vv --backend-flags=--use-merge |& grep 'generated tests'
+///
+/// should result in
+///
+///       test_original: {"generated tests": 1024, "total instructions": 709566, "completed paths": 4093}
+///       test_merged: {"generated tests": 1, "completed paths": 41, "total instructions": 11852}
+///
+/// Indicating that the original version suffers from a path explosion and generates
+/// more tests, executes more instructions and explores more paths than the merged version.
+
+use verification_annotations as verifier;
+
+fn main() {
+    println!("Hello, world!");
+}
+
+const N: usize = 10;
+
+/// Test the impact of not using path merging.
+///
+/// The critical part of this example is the second loop
+/// that contains multiple branches where the branches depend
+/// on symbolic expressions.
+///
+/// On a classic symbolic execution tool, this will result in
+/// 2^N paths.
+#[test]
+fn test_original() {
+    // An array
+    let mut a = [0u32; N];
+
+    // Set each element of array to a symbolic value
+    for i in &mut a {
+        *i = verifier::AbstractValue::abstract_value();
+    }
+
+    // A loop containing two branches - this will cause a performance problem
+    // for conventional symbolic execution.
+    for x in &a {
+        verifier::assume((5..10).contains(x) || (15..20).contains(x))
+    }
+
+    // A true assertion about an arbitrary element of the array
+    verifier::assert!(a[3] < 20);
+}
+
+/// Test the impact of using path merging.
+///
+/// This test reduces the number of paths and instructions executed by merging
+/// all the paths forked inside the loop.
+///
+/// There should be just one path when verified using KLEE with --use-merge
+#[test]
+
+/// The test depends on KLEE-specific features
+#[cfg(feature = "verifier-klee")]
+fn test_merged() {
+    // An array
+    let mut a = [0u32; N];
+
+    // Set each element of array to a symbolic value
+    for i in 0..a.len() {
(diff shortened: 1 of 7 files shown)


demos/bottlenecks/merging/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+///
+/// On a classic symbolic execution tool, this will result in
+/// 2^N paths.
+#[test]
+fn test_original() {
+	// An array
+    let mut a = [0u32; N];
+
+	// Set each element of array to a symbolic value
+	for i in 0..a.len() {
```
fshaked: Alternatively:

```rust
for i in &mut a {
  *i = verifier::AbstractValue::abstract_value();
}
```

demos/bottlenecks/merging/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+		a[i] = verifier::AbstractValue::abstract_value();
+	}
+
+	// A loop containing two branches - this will cause a performance problem
+	// for conventional symbolic execution.
+	for x in a.into_iter() {
```
fshaked: rustc gives a warning for this use of `into_iter`. use `iter` instead, or even better use `for x in &a {`

demos/bottlenecks/merging/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+		verifier::assume((5..10).contains(x) || (15..20).contains(x))
```
fshaked: Wouldn't it be more realistic to create the branching using rust primitives?

verification-annotations/src/klee.rs
```diff
@@ -22,12 +22,68 @@ extern "C" {
...
+unsafe fn get_value_u32(x: u32) -> u32 {
+    klee_get_value_i32(x as i32) as u32
+}
+
+unsafe fn get_value_u64(x: u64) -> u64 {
+    klee_get_value_i64(x as i64) as u64
+}
+
+unsafe fn get_value_i128(x: i128) -> i128 {
+    let x = x as u128;
```
fshaked: This looks like a typo, you probably meant `get_value_u128(x as u128) as i128`

fshaked (changes requested):
The into_iter needs to be changed.

demos/bottlenecks/merging/src/main.rs
alastairreid: The branching is inside contains() and from use of || - so, pedantically, the branching does come from Rust primitives.
But you're probably wanting me to avoid verifier::assume().
Everything I can think of to do that makes the code longer and more complex or stops it from exploding - one critical feature of the code is that the exploding loop must depend on symbolic data (or it won't explode).
I'm happy to update this as we gain experience and find other examples in the future.

verification-annotations/src/klee.rs
alastairreid: I wasn't sure whether I should call get_value_u128 as you suggest or cut and paste the definition and add appropriate casts.
In the end, I decided to go with cut and paste so that it is more standalone.

But I think you're right - the shorter version is easier to read and there is no obvious down-side.
So I'll make the change - thanks.

fshaked:
I checked out the merge branch and tried to run the following (in docker):
```
cd demos/bottlenecks/merging
cargo verify --backend=klee --tests -vv --backend-flags=--use-merge
```
`cargo build` fails with this error:
```
INFO - STDERR: error: could not copy "/home/sflur/workspace/rust-verification/rust-verification-tools/demos/bottlenecks/merging/target/x86_64-unknown-linux-gnu/debug/deps/verification_annotations-d73fc9eef3bc265c.4ul91e98bals9drv.rcgu.bc" to "/home/sflur/workspace/rust-verification/rust-verification-tools/demos/bottlenecks/merging/target/x86_64-unknown-linux-gnu/debug/deps/verification_annotations-d73fc9eef3bc265c.bc": No such file or directory (os error 2)                                     
```
I see that a similar .o file exists.

alastairreid:
Yeah, the problem is something to do with
verification_annotations/lib/klee.c (that I added to define
klee_is_symbolic).

With the file, demos/bottlenecks/merging fails to link.
Without the file, verificaiton_annotations/src/test.rs fails to link.

I'm trying to figure out why.

A

alastairreid:
I fixed the linking problem by changing cargo-verify to define the symbol klee_is_symbolic instead of using verification_annotations/build.rs (etc) to create a definition that we never use (and that caused the linking issues).

And I made sure that scripts/regression-tests ran ok (for KLEE).

In response to your review I've changed everything except the coherent! demo/test which still uses assume, contains and || in the exploding loop. (I don't have a good, short alternative, and I think it is ok as it is.)

PR #108, opened 2021-03-08
fix(demo): change deps path to be relative

priyasiddharth:


priyasiddharth:
@fshaked PTAL.

alastairreid (approved):
We have been using absolute paths in documentation because (I think that) it is a bit easier to understand how to adapt it for your own code.

But I think it is ok to use relative paths in almost(*) all of the actual working demos.

I'm not sure whether we should pre-emptively change (almost) all of the demos now or to do it on demand?

* "almost" because having one example that matches the documentation is useful to detect when the documentation needs updated.

PR #110, opened 2021-03-11
Distinguish different kinds of error

alastairreid:
This change makes finer distinctions in cargo-verify between different kinds of error such as assertion failure, array index bounds failure, and overflow.


demos/simple/errors/verify.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+cargo clean
+
+# this should detect multiple errors
+( cargo-verify --tests --verbose > out || true )
```
fshaked: You can use `| tee out` instead of `> out` and `cat out`.

Also, I don't think you really mean to run this line in a subshell, so remove the `(` `)`.

demos/simple/errors/verify.sh
```diff
@@ -0,0 +1,20 @@
...
+cat out
+
+# check for all expected errors
+grep -q "test assert_eq_should_fail ... ASSERT_FAILED" out
```
fshaked: add `-F` to interpret the string as fixed text instead of the default regular expression.

fshaked (approved):

demos/simple/errors/verify.sh
alastairreid: cool - done

demos/simple/errors/verify.sh
alastairreid: good idea - done

PR #113, opened 2021-03-17
cargo-verify: tweak verbosity

alastairreid:
This incomplete change causes commands to only print stdout/stderr if their level
is below the current verbosity.

To finish this off, I (Alastair) plan to add -q/--quiet support as well.

### DIFF
--- a/cargo-verify/src/main.rs
+++ b/cargo-verify/src/main.rs
@@ -474,7 +474,7 @@ fn build(opt: &Opt, package: &str, target: &str) -> CVResult<PathBuf> {
         .arg(runtime)
         .arg(&bc_file)
         .args(&c_files)
-        .output_info(&opt)?;
+        .output_info(&opt, 3)?;
     bc_file = new_bc_file;
 
     if opt.backend == Backend::Seahorn {
@@ -572,7 +572,7 @@ fn compile(opt: &Opt, package: &str, target: &str) -> CVResult<(PathBuf, Vec<Pat
     cmd.arg(format!("--target={}", target))
         .args(vec!["-v"; opt.verbose])
         .envs(get_build_envs(&opt)?)
-        .output_info(&opt)?;
+        .output_info(&opt, 0)?;
     // .env("PATH", ...)
 
     // Find the target directory
@@ -613,7 +613,7 @@ fn compile(opt: &Opt, package: &str, target: &str) -> CVResult<(PathBuf, Vec<Pat
             if opt.tests || !opt.test.is_empty() {
                 Err("  FAILED: Use --tests with library crates")?
             } else {
-                Err(format!("  FAILED: Test {} compilation error", &package))?
+                Err(format!("  FAILED: Test {} unable to find the right bitcode file - should you have used --tests?", &package))?
             }
         }
         _ => {
@@ -664,7 +664,7 @@ fn patch_llvm(opt: &Opt, options: &[&str], bcfile: &Path, new_bcfile: &Path) -> 
         .arg(new_bcfile)
         .args(options)
         .args(vec!["-v"; opt.verbose])
-        .output_info(&opt)?;
+        .output_info(&opt, 3)?;
     Ok(())
 }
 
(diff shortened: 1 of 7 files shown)


cargo-verify/src/main.rs
```diff
@@ -274,6 +279,9 @@ fn process_command_line() -> CVResult<Opt> {
     // Use the user specified number of jobs, or the number of CPUs.
     opt.jobs = opt.jobs_arg.unwrap_or(num_cpus::get());
 
+    // Normal mode is verbosity 1, quiet is 0
+    opt.verbose = if opt.quiet { 0 } else { opt.verbose + 1 };
```
fshaked: This should be done earlier (before the use of `info_at!` above).

cargo-verify/src/main.rs
```diff
@@ -413,13 +421,13 @@ fn verify(opt: &Opt, package: &str, target: &str) -> CVResult<Status> {
     let end = Instant::now();
 
     println!(
-        "test result: {:#}. {} passed; {} failed",
+        "\ntest result: {:#}. {} passed; {} failed",
```
fshaked, line 482: The `\n` at the beginning needs to be documented. I'm guessing it's here to close the `print!` from `verifier_run` in the -q case?

cargo-verify/src/main.rs
```diff
@@ -438,7 +446,23 @@ fn verifier_run(opt: &Opt, bcfile: &Path, name: &str, entry: &str) -> Status {
         Status::Unknown
     });
 
-    println!("test {} ... {:#}", name, status);
+    if opt.quiet {
+        let s = match status {
+            Status::Unknown => Colour::Yellow.paint("?"),
+            Status::Verified => Colour::Green.paint("."),
+            Status::Error => Colour::Red.paint("F"),
```
fshaked, line 512: Maybe E instead of F?

cargo-verify/src/main.rs
```diff
@@ -438,7 +446,23 @@ fn verifier_run(opt: &Opt, bcfile: &Path, name: &str, entry: &str) -> Status {
...
+            Status::Overflow => Colour::Red.paint("O"),
+            Status::Panic => Colour::Red.paint("P"),
+            Status::Reachable => Colour::Red.paint("R"),
+            Status::Timeout => Colour::Red.paint("T"),
+        };
+        print!("{}", s);
+    } else {
+        println!("test {} ... {:#}", name, status);
+    }
+    term.flush();
```
fshaked: error[E0425]: cannot find value `term` in this scope

fshaked (changes requested):

fshaked:
I don't like that the colours can't be switched off. This makes greping harder and redirected files look bad. It doesn't seem like ansi-term supports switching the colours off. I find it very surprising considering how popular this crate is.

cargo-verify/src/main.rs
alastairreid: Good catch

cargo-verify/src/main.rs
alastairreid: I chose F and . (and the colours) to match cargo-test.

cargo-verify/src/main.rs
alastairreid: Fixed. (I'm not sure if this was a version issue - because I remember seeing that this fixed the problem (of no output until all tests completed)

cargo-verify/src/main.rs
alastairreid: Good idea - done

alastairreid:
oops, missed your comment about turning colours off.
Maybe there's another crate that supports it?

(Also, you will notice that I replaced all those magic verbosity numbers with an enum - hopefully easier to understand as we look at the code.)

fshaked:
There's yansi that can enable and disable colours on the fly, but it can't
check for terminal features (a bit old and not so popular?). You can use
https://docs.rs/console/0.14.1/console/ to check for features, and it also
has colour support.

PR #116, opened 2021-03-18
Docker: build libstd/libcore without vectorization

alastairreid:
This fixes linking errors we were getting due to compiling libstd/libcore
with vectorization enabled and compiling code we are verifying
with vectorization disabled.

### DIFF
--- a/docker/rustc/Dockerfile
+++ b/docker/rustc/Dockerfile
@@ -7,7 +7,7 @@ WORKDIR ${USER_HOME}
 
 # Fetch rustc
 ENV RUSTC_DIR=${USER_HOME}/rust
-ARG RUSTFLAGS_STAGE_NOT_0="-Cembed-bitcode=yes"
+ARG RUSTFLAGS_STAGE_NOT_0="-Cembed-bitcode=yes -Ctarget-feature=-mmx,-sse,-sse2,-sse3,-ssse3,-sse4.1,-sse4.2,-3dnow,-3dnowa,-avx,-avx2"
 ARG RUSTC_VERSION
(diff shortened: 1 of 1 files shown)


alastairreid:
Sadly, this doesn't seem to help.
Which is a shame because it takes hours to rebuild it...

alastairreid:
What I have learned:

- Building libcore with -sse2 disabled results in ld.lld errors about SSE2 register returns with SSE2 disabled

  > ld.lld: error: /home/adreid/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-2.33.3/src/suggestions.rs:22:26: in function _ZN4clap11suggestions24did_you_mean_flag_suffix28_$u7b$$u7b$closure$u7d$$u7d$17h4e49fad07f2d71f3E void (%"core::option::Option<(usize, (alloc::string::String, core::option::Option<&str>))>"*, { i64*, i64* }*, %"clap::app::App"*): SSE2 register return with SSE2 disabled

- Enabling SSE2 in both libcore/libstd and when compiling normal crates (e.g., hashbrown, encoding-rs, etc.) avoids compilation/linking errors but results in bitcode files with hand-vectorized code in them. These cause KLEE to abort exploring paths with a warning message about pmovmskb. Inspecting the files, it is clear that this is just a common *first* instruction in a sequence and we would have to implement many more of these instructions in KLEE or rvt-patch-llvm.

- Inspecting the code of libraries like hashbrown and vectorization is normally enabled on x86 and x86_64 but disabled with miri - so maybe enabling the miri feature would disable the hand-vectorized code?

alastairreid:
This has been replaced by #125 (which, unfortunately doesn't work)

PR #118, opened 2021-03-19
Add --replace-backend-flags to cargo-verify

fshaked:
Also add Handlebar templates to --backend-flags.

### DIFF
--- a/cargo-verify/src/seahorn.rs
+++ b/cargo-verify/src/seahorn.rs
@@ -78,39 +78,46 @@ fn importance(line: &str, expect: &Option<&str>, name: &str) -> i8 {
 /// Run Seahorn and analyse its output.
 fn run(opt: &Opt, name: &str, entry: &str, bcfile: &Path, out_dir: &Path) -> CVResult<Status> {
     let mut cmd = Command::new("sea");
-    cmd.args(&["bpf",
-               // The following was extracted from `sea yama -y VCC/seahorn/sea_base.yaml`
-               "-O3",
-               "--inline",
-               "--enable-loop-idiom",
-               "--enable-indvar",
-               "--no-lower-gv-init-struct",
-               "--externalize-addr-taken-functions",
-               "--no-kill-vaarg",
-               "--with-arith-overflow=true",
-               "--horn-unify-assumes=true",
-               "--horn-gsa",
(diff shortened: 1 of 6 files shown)


fshaked:
@alastairreid, note that --replace-backend-flags doesn't pass `Opt::args` to the verified program, because there is no simple way to do it (will need a handlebar template that expands to Vec<String>, or something like that). The user should pass those through --backend-flags instead, if she uses --replace-backend-flags.

alastairreid:
Just before I dive into the code...

Passing args to the program is a really useful capability - it's fundamental for finding bugs in standalone programs like uutils/coreutils.
So I would **really** not like to lose that capability.

cargo-verify/src/seahorn.rs
```diff
@@ -78,39 +78,46 @@ fn importance(line: &str, expect: &Option<&str>, name: &str) -> i8 {
...
-        .args(&opt.backend_flags)
-        .arg(&bcfile);
-    // .args(&opt.args)
+
+    let user_flags: Vec<_> = opt.backend_flags.iter().map(|flag| {
+        backends_common::format_flag(&flag, &entry, &bcfile, &out_dir)
+    }).collect::<Result<_,_>>()?;
+
+    if ! opt.replace_backend_flags {
+        cmd.args(&["bpf",
```
alastairreid, line 87: I thought that the plan was to be able to provide this content as a yaml file?

alastairreid (commented):
My main thought when reviewing a change like this is "how does this change the user interface?"
So it would be good if this change also included tests or changes to existing tests that show the necessary changes.

alastairreid:
Following on from my previous comment...
If this changes how arguments are passed to programs, does this require a change to scripts/regression-tests or do those tests (especially of demos/simple/argv) still pass?

cargo-verify/src/seahorn.rs
fshaked: As you already noted in #119, this was implemented separately.

fshaked:
> If this changes how arguments are passed to programs, does this require a change to scripts/regression-tests or do those tests (especially of demos/simple/argv) still pass?

This PR doesn't change anything if you don't use `--replace-backend-flags` on the command line. So script/regression-tests should be unaffected.

I will add to regression-tests something that uses `--replace-backend-flags`.

PR #119, opened 2021-03-19
Add --verify_common_dir option to cargo-verify

fshaked:


cargo-verify/src/seahorn.rs
```diff
@@ -77,40 +77,21 @@ fn importance(line: &str, expect: &Option<&str>, name: &str) -> i8 {
 
 /// Run Seahorn and analyse its output.
 fn run(opt: &Opt, name: &str, entry: &str, bcfile: &Path, out_dir: &Path) -> CVResult<Status> {
+    let verify_common_dir = match &opt.verify_common_dir {
+        Some(verify_common_dir) => verify_common_dir,
+        None => Err("The '--verify_common_dir' option is missing")?,
+    };
+
     let mut cmd = Command::new("sea");
-    cmd.args(&["bpf",
```
alastairreid: Ah, this looks like what I was asking for in the other, separated commit!
Moving this content into an external file.

docker/seahorn/Dockerfile
```diff
@@ -1,16 +1,20 @@
...
 ARG SEAHORN_VERSION
 ARG YICES_VERSION
 
+# cargo-verify relies on this variable to find the yaml files
 ENV VERIFY_COMMON_DIR=/home/seahorn/verify-c-common
 
-RUN git clone --no-checkout https://github.com/yvizel/verify-c-common.git ${VERIFY_COMMON_DIR} \
+RUN git clone --no-checkout https://github.com/seahorn/verify-c-common.git ${VERIFY_COMMON_DIR} \
   && cd ${VERIFY_COMMON_DIR} \
   && git checkout ${VERIFY_COMMON_VERSION}
```
alastairreid: Can you change the name of this environment variable to include the word "SEAHORN"?
Ditto for adding the command line argument to cargo-verify.

As it is, the directory does not look very "common" since it is SeaHorn specific and it's not obvious how we would generalize it for KLEE, SMACK, Crux or any other tool that we want to pass configuration files to.

fshaked: Done

alastairreid (approved):
LGTM

PR #122, opened 2021-03-29
Minor fixes to Using KLEE

fshaked:


docs/_posts/2020-09-01-using-klee.md
```diff
@@ -64,8 +64,8 @@ The Rust compiler and KLEE are in the Dockerfile so start the Docker image
 by running
 
 ``` shell
-cd demos/simple/klee
```
alastairreid, line 67: Both work - and the former is the way I tend to run it.

I think the existing order is better because it shows what you would do if verifying Oak, for example.

```
cd $HOME/oak # or wherever it is
$RVT_DIR/docker/run
# because we mount $PWD in docker, you are now in Oak and can access all the Oak files.
```

docs/_posts/2020-09-01-using-klee.md
```diff
@@ -27,7 +27,7 @@ and the `cargo-verify` script
 as described [here][Using PropVerify].
 This note describes how to use the KLEE directly
 in case you wonder how `cargo-verify` works or want
-to add support for a different tool.6
```
alastairreid, line 30: oops, good catch!

docs/_posts/2020-09-01-using-klee.md
fshaked: I don't think the current one works. After you cd to klee docker/run is an invalid command. Am I missing something?

PR #123, opened 2021-03-29
Add a short README to docs

fshaked:


docs/README.md
```diff
@@ -0,0 +1,21 @@
+# The rust-verificatio-tools blog
```
alastairreid: Cool, thanks for writing this!

Can you move it to docs/_internal/howto-blog.md
so that it doesn't accidentally become part of the website (the _ part)
it is clear that it is documentation for us (the internal part)
and it's a bit easier to guess what it is about from the filename.

docs/README.md
```diff
@@ -0,0 +1,21 @@
...
+gem install jekyll bundler --user-install
+bundle config set --local path ~/.local/share/gem/
+bundle install
+```
+
+Check that the blog builds and looks ok:
+
+``` shell
+# cd docs
+bundle exec jekyll serve
```
alastairreid: I tend to use --livereload --incremental (or -L -I for short) to:
- only rebuild pages that changed (faster)
- update the web browser automatically on every change (no need to ctrl-R to reload)

alastairreid: Also --open-url (-o for short) will automatically open the browser for you.
So the complete invocation is 

> bundle exec jekyll serve -o -I -L

docs/README.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+and point your browser to [here](http://127.0.0.1:4000/rust-verification-tools/).
```
alastairreid: not needed if you use -o

docs/README.md
fshaked: Done

docs/README.md
fshaked: (There is no -L option, I assume you meant -l)
Done.

docs/README.md
fshaked: Removed

PR #124, opened 2021-03-31
Add --output option to cargo-verify

alastairreid:


cargo-verify/src/main.rs
```diff
@@ -395,6 +404,14 @@ fn verify(opt: &Opt, package: &str, target: &str) -> CVResult<Status> {
     info_at!(&opt, Verbosity::Informative, "  Building {} for verification", package);
     let bcfile = build(&opt, &package, &target)?;
 
+    info_at!(&opt, Verbosity::Informative, "  Generated binary {}", bcfile.to_string_lossy());
```
fshaked: In the `structopt` comment for `output` you use "LLVM bitcode file", I think that is better than "binary" here (and a few lines below).

cargo-verify/src/main.rs
```diff
@@ -395,6 +404,14 @@ fn verify(opt: &Opt, package: &str, target: &str) -> CVResult<Status> {
...
+
+    if let Some(output) = &opt.output {
+        std::fs::copy(bcfile, output)?;
+        info_at!(&opt, Verbosity::Informative, "Wrote binary to {}", output.to_string_lossy());
+        return Ok(Status::Verified)
```
fshaked: Maybe use `exit(0)` instead of `return`? This will avoid printing the "VERIFICATION_RESULT: {}" which is misleading.

fshaked (changes requested):

fshaked:
Another thing, the new `output` option has no effect when using proptest. Either make it an illegal combination (report an error), or implement it for proptest.

cargo-verify/src/main.rs
alastairreid: Good idea - done

cargo-verify/src/main.rs
alastairreid: Good idea - done

PR #125, opened 2021-03-31
Fix sse2 problem

alastairreid:


alastairreid:
I forgot to run the regression tests before pushing this.
Unfortunately, it fails almost all regression tests because it causes linking errors.
I will need to revisit this and try again.

PR #127, opened 2021-04-19
Simd support

alastairreid:
Add SIMD emulation library that emulates x86 SIMD intrinsic operations.
This emulation library is used to replace calls to the intrinsics in the bitcode file (in rvt_patch_llvm).


rvt-patch-llvm/src/main.rs
```diff
@@ -52,6 +54,10 @@ struct Opt {
     #[structopt(short, long)]
     initializers: bool,
 
+    /// Substitute function calls for SIMD intrinsic operations
+    #[structopt(short, long)]
+    intrinsics: bool,
```
fshaked, line 58: The short form `i` is already used by `initializers`, this causes `rvt-patch-llvm --help` to panic.
(maybe add a regression test?)

rvt-patch-llvm/src/main.rs
```diff
@@ -101,6 +107,29 @@ fn main() {
         handle_initializers(&context, &mut module);
     }
 
+    if opt.intrinsics {
+        // intrinsics
+        let is = get_function(&module, |name| name.starts_with("llvm.x86"));
+
+        // simd_emulation functions
+        let rs = get_function(&module, |name| name.starts_with("llvm_x86"));
+        let rs: HashMap<&str, FunctionValue> = HashMap::from_iter(rs.iter().filter_map(|f| f.get_name().to_str().ok().map(|nm| (nm, *f))));
```
fshaked: I think it's nicer to use `collect`, instead of `from_iter`, here:
 rs.iter().filter_map(|f| f.get_name().to_str().ok().map(|nm| (nm, *f))).collect();

simd_emulation/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+LINK = llvm-link-10
+
+RS_SRCS := $(wildcard src/*.rs)
+RS_OBJS := $(patsubst src/%.rs, build/%.bc, $(RS_SRCS))
+
+OBJS := $(RS_OBJS)
+
+build/%.bc: src/%.rs
+	mkdir -p build
```
fshaked: Suggestion:
mkdir -p $(dir $@)

simd_emulation/Makefile
```diff
@@ -0,0 +1,35 @@
...
+	$(RUSTC) $(RSFLAGS) $^ -o $@
+
+simd_emulation.bc: $(OBJS)
+	$(LINK) $(OBJS) -o $@
+
+clean::
+	$(RM) $(OBJS)
```
fshaked: Note that if you rename/remove one of the src/*.rs files its coresponding .bc file will not be cleaned. Suggestion:
$(RM) build/

simd_emulation/src/lib.rs
```diff
@@ -0,0 +1,1234 @@
...
+/// using 2nd order functions
+mod vector {
+
+    /// Implemented by types that support 4-element vectors
+    /// Provides methods to construct, destruct and convert vectors to their
+    /// native representation.
+    pub trait Vector2
+    where
+        Self: Copy,
+        Self::Vec: Copy,
```
fshaked: This is a bit confusing, wouldn't it be better to move this copy to the line below:
`type Vec: copy;`
I think this has the same meaning but is much more clear.

simd_emulation/src/lib.rs
```diff
@@ -0,0 +1,1234 @@
...
+        fn get25(x: Self::Vec) -> Self;
+        fn get26(x: Self::Vec) -> Self;
+        fn get27(x: Self::Vec) -> Self;
+        fn get28(x: Self::Vec) -> Self;
+        fn get29(x: Self::Vec) -> Self;
+        fn get30(x: Self::Vec) -> Self;
+        fn get31(x: Self::Vec) -> Self;
+    }
+
+    /// Define Vector4 implementation for vectors of $ety and constructor $c
```
fshaked: `$c`? 
Maybe `$pty`?

simd_emulation/src/lib.rs
```diff
@@ -0,0 +1,1234 @@
...
+            pub struct $pty(pub $ety, pub $ety);
+
+            impl Vector2 for $ety {
+                type Vec = $pty;
+
+                fn new(x0: Self, x1: Self) -> Self::Vec {
+                    $pty(x0, x1)
+                }
+
+                fn get0(x: Self::Vec) -> Self {
```
fshaked: 1. I think `from_vec0` will be a better name.
2. I think the function can take a reference: `&Self::Vec` and then you don't need the Copy constraint on `type Vec`.

simd_emulation/src/lib.rs
```diff
@@ -0,0 +1,1234 @@
...
+    pub fn cmpeq_u16(x: u16, y: u16) -> u16 {
+        if x == y {
+            0xffff
+        } else {
+            0x0
+        }
+    }
+
+    pub fn cmpgt_u8(x: u8, y: u8) -> u8 {
+        if x == y {
```
fshaked: From the `gt` in the name I would expect `>` instead of `==`.

simd_emulation/Makefile
```diff
@@ -0,0 +1,35 @@
...
+# Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+# http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+# <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+# option. This file may not be copied, modified, or distributed
+# except according to those terms.
+
+# this rule comes first to make sure that it is the default target
+default: simd_emulation.bc
+
+RUSTC = rustc
```
fshaked: Use `?=` instead of `=`, so users can set it in their shell init script.

simd_emulation/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
fshaked, line 1: I think the intention is to build this with make, so remove the Cargo.toml file.

cargo-verify/src/main.rs
```diff
@@ -565,12 +565,15 @@ fn build(opt: &Opt, package: &str, target: &str) -> CVResult<PathBuf> {
     let backend_lc = format!("{}", opt.backend).to_lowercase();
     let runtime = format!("{}/runtime/rvt-{}.bc", rvt_dir, backend_lc);
     let runtime = PathBuf::from(&runtime);
+    let simd_emulation = format!("{}/simd_emulation/simd_emulation.bc", rvt_dir);
+    let simd_emulation = PathBuf::from(&simd_emulation);
```
fshaked: Everything from `let rvt` above, down to this line (including) can be rewritten more idiomatically like this:
```rust
    let rvt_dir = PathBuf::from( std::env::var("RVT_DIR")? );
    let runtime = rvt_dir.clone()
        .append("runtime")
        .append(format!("rvt-{}.bc", opt.backend.to_string().to_lowercase()));
    let simd_emulation = rvt_dir.clone()
        .append("simd_emulation")
        .append("simd_emulation.bc");
```

fshaked (changes requested):

rvt-patch-llvm/src/main.rs
alastairreid: Proposed fix: remove the short form from the following flags: features, initializers, seahorn, intrinsics.
Having cryptic sequences of flags is not especially useful.
(Having "-o", "-v", etc. where the short form is conventional is fine - will leave them.)

simd_emulation/Makefile
alastairreid: Done

rvt-patch-llvm/src/main.rs
alastairreid: Nice - done

simd_emulation/src/lib.rs
alastairreid: oops
Though, on closer inspection, I noticed that the function was dead code and deleted it.
(I used to define more intrinsics - but then I noticed that LLVM was turning them into generic (architecture independent) intrinsics so I did not need to define them after all.)

simd_emulation/Makefile
alastairreid: Done

simd_emulation/Cargo.toml
alastairreid: I'd like to be able to share the code with miri - and they want a Cargo file.
So keeping both.

(I may end up abandoning the Makefile at some point. It only works well as long as the entire crate is one file. But it makes it easy to find the bitcode file reliably afterwards so I like it for now.)

cargo-verify/src/main.rs
alastairreid: Done

simd_emulation/Makefile
alastairreid: Good catch - done

simd_emulation/src/lib.rs
alastairreid: Done

simd_emulation/src/lib.rs
alastairreid: I'm not sure from_vec0 is better (or worse) than get0 - inclined to leave that as is for now.
I've added the & to definitions and calls.

simd_emulation/src/lib.rs
alastairreid: Your idea of adding & on the get arguments was a better fix.

fshaked (approved):

PR #128, opened 2021-05-05
Refactor verification-annotations

fshaked:
(this is a bit of a big one, sorry)

Added a prelude to verification-annotations to make it easier to import.

Added verifier module, that re-exports klee/seahorn, and moved shared code
from those modules to verifier.

Reimplemented verifier_nondet_bytes (not using klee_make_symbolic anymore).

Added a string module to propverify with strategies for generating an arbitrary
string with `n` bytes (doesn't work very well), and ascii string with `n`
bytes (works quite well with Klee).

Added demo/simple/string to test the new string module.


alastairreid (approved):
Awesome!
Thanks for doing this.

PR #132, opened 2021-05-07
CI and rustfmt

alsuren:
Contains:

- [x] A basic rustfmt --check github action.
- [X] A commit to fix everything up.

You can see what the build output would look like for real here: https://github.com/alsuren/rust-verification-tools/pull/1

I've kept it super-simple for now, because I'm not expecting anything more complex to work without a working docker environment to run tests in.

It turns out we're pinned to nightly from 1.46.0 for building things in the docker containers, but we're using nightly rustfmt features that are not supported that far back, and also features that are deprecated on today's nightly. I have decided to arbitrarily go with today's nightly and ignore the deprecation warnings for now.

It's probably best to merge #128 first, because re-running rustfmt will be easier than fixing conflicts on that PR.


alastairreid:
This looks great - thank you for that.
Now that #128 has been merged could you update it?

(We could also update it if you prefer - the CI integration bit is the really valuable bit where you have helped us since we have not set that kind of thing up before.)

alsuren:
Rebased and force-pushed.

Once I have a working set of docker images (I will probably just comment out that failing test or something), I will try to push them to a registry and make a PR that uses them to run `cargo test` in a couple of the sub-projects on push (as a separate PR). I suspect that this will be a bit more tricky than this rustfmt integration, but hopefully it will still be understandable once it's finished.

alastairreid (approved):

Issue #131, opened 2021-05-07
docker/build script fails to build klee


alsuren:
Full output lives here: [docker-build-output.txt](https://github.com/project-oak/rust-verification-tools/files/6442436/docker-build-output.txt) and I have included a hopefully-relevant snippet below.

I don't have any experience with klee, so I haven't tried digging into this too deeply.

I did notice that we have UCLIBC_VERSION defined as `klee_uclibc_v1.2` in `docker/mkimage` but it's not used anywhere. There are about 6 commits between `https://github.com/klee/klee-uclibc.git` `HEAD` and `klee_uclibc_v1.2`, so my current theory is that it might be one of those.

I had also removed the sudo around the docker calls, because my user already has permission to use docker (on macos). I'm currently doing a full-rebuild with sudo enabled and `git checkout UCLIBC_VERSION` in `docker/klee/build_klee`. I will tell you how it goes.

I do wonder whether it would be worthwhile publishing the relevant docker images somewhere. This would also help with https://github.com/project-oak/rust-verification-tools/issues/2 , and it's a workflow that people will be used to if they have used `cross` before.

```
 > [9/9] RUN sh build_klee:                                                                                                        
#14 0.331 Cloning into 'klee-uclibc'...                                                                                            
#14 4.260 INFO:Forcing C compiler to be.../usr/bin/clang-10
#14 4.260 INFO:Absolute path to compiler.../usr/bin/clang-10
#14 4.260 INFO:Disabling assertions
#14 4.260 INFO:Configuring for Debug build
#14 4.260 INFO:Configuring for LLVM bitcode archive
...
#14 161.5 [ 93%] Built target RuntimeFortify
#14 161.5 [ 93%] Built target BuildKLEERuntimes
#14 161.6 [ 94%] Built target kleeModule
#14 161.6 [ 97%] Built target kleeCore
#14 161.7 [ 97%] Built target klee
#14 161.7 Scanning dependencies of target systemtests
#14 161.7 [ 97%] Running system tests
#14 161.8 -- Testing: 346 tests, 4 workers --
#14 161.8 Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60..
#14 185.0 FAIL: KLEE :: Runtime/POSIX/FD_Fail.c (241 of 346)
#14 185.0 ******************** TEST 'KLEE :: Runtime/POSIX/FD_Fail.c' FAILED ********************
#14 185.0 Script:
#14 185.0 --
[...]

PR #133, opened 2021-05-07
pin down klee UCLIBC_VERSION

alsuren:
This doesn't solve the problem I'm experiencing in #131 but it feels like good practice, so I'm submitting it anyway. Feel free to close.


alastairreid (approved):
Thanks for this.

We clearly meant to do this... but dropped the ball. So really happy to have this fix!

PR #135, opened 2021-05-18
Docker cleanup

alastairreid:
Faster, simpler docker build with initial steps towards LLVM11 support.

- We no longer build rustc/libstd/libcore/etc.
  This reduces docker build time by several hours
  and dramatically reduces image size.
- Use docker/init to prebuild tools and libraries
- Merge docker/{minisat,stp,yices,z3} -> docker/solvers
- All tools are built in subdirectories of $USERHOME by $USERNAME
  instead of tool $T being built in /home/$T by root
- Initial work towards supporting LLVM11
  - the LLVM version is now explicit in all tools/etc. (defaults to 10)
  - LLVM11 support works for 'cargo verify' but not for 'cargo verify --tests'
    This causes most of the regression tests to fail.
    So this support is only usable in limited (but useful) cases


.dockerignore
```diff
@@ -0,0 +1,45 @@
+/.git
+/bin
+/docs
+
+/downloads
+/experiments
+/runtime/build_*
+
+# .dockerignore files differ from .gitignore files in that
```
fshaked, line 9: You can use `**`, e.g., `**/*.bc` will ignore `*.bc` in all levels (including root).

docker/mkimage
```diff
@@ -15,6 +15,18 @@ readonly DOCKER_UID="${UID:-0}"
 readonly DOCKER_GID="$(id -g)"
 readonly DOCKER_USER="${USER:-root}"
 
+if [ -v LLVM11 ]; then
```
fshaked, line 18: I think the use of `-v` is discouraged, use `-n "$LLVM11"` instead.

cargo-verify/src/main.rs
```diff
@@ -97,6 +97,15 @@ pub struct Opt {
     #[structopt(long, value_name = "PATH", env = "SEAHORN_VERIFY_C_COMMON_DIR")]
     seahorn_verify_c_common_dir: Option<String>,
 
+    /// Which LLVM version to use (e.g., 10 or 11)
```
fshaked, line 100: Shouldn't the version of llvm be fixed at compile time (of cargo-verify, assuming it's the same llvm version that rvt-patch-llvm and runtime is build with).
The runtime switch is a bit misleading as it has no effect on the version of llvm baked into our runtime libraries and rvt-patch-llvm, which will probably lead to hard to understand errors when things fail.
I'm not sure what is the best way to deal with this.

fshaked (changes requested):
Also, I think instead of docker/init each component should have it's own Dockerfile in the relevant source folder, which will copy only that source and build it, instead of copying the whole repo.
(but we don't have to do this now)

cargo-verify/src/main.rs
alastairreid: The question is: how to bake the version in?
This allows it to be set by an environment variable (that is set by the docker container) or to be set by a command line.
I think this is pretty flexible and convenient.

(The other alternatives I can think of are to generate a .rs file that is used at build time to bake the version in or to generate a configuration file that is read when cargo-verify starts. I think both of those are less convenient.)

.dockerignore
alastairreid: Cool - will do

cargo-verify/src/main.rs
fshaked: No, this runtime switch does not affect the build of rvt-patch-llvm or the build of the code in runtime/src (you forgot to change the Makefile in the latter? it uses clang-10), if you build those with llvm-11 and then run cargo-verify with llvm-10 you will probably run into trouble.

alastairreid: Don't forget to

1) run docker/init which uses LLVM_VERSION when rebuilding rvt_patch_llvm
2) make an LLVM11-based rustc your default in rustup. I am using nightly-2021-02-20-x86_64-unknown-linux-gnu (which is what Rust-for-Linux depends on).

If you do this, I believe that everything will be consistent.
Please let me know if I missed anything.

fshaked: You can indeed get things to be consistent, but you're making it unnecessarily easy to get things inconsistent. LLVM version is not a runtime option for any other tool (external or internal to the project) and it shouldn't be for cargo-verify.

alastairreid: I think this PR works and I have not seen problems.
Since we are currently only recommending LLVM10 at the moment and that is the default, it should not cause problems for users either.
If there is an easy improvement we could make, I'm open to suggestions - but let's not overcomplicate this.

docker/mkimage
alastairreid: I think it is discouraged only because old versions of bash do not support -v.
Since this is part of the docker build process, I'm happy to assume that people have an up to date machine.

Issue #139, opened 2021-05-20
Why sudo required for building image?


gww-parity:
Why do you require sudo to build docker images?

I would love to test Klee but requireemnt to run `docker build` with sudo makes me suspicious.

gww-parity:
Can we make some workaround to not require `sudo`?

alastairreid:
On a Mac, you don't need sudo and, if you are used to managing packages with homebrew (which explicitly avoids sudo) then you will be disturbed by having to use sudo.

But, if you are on Linux then you need to use sudo with docker (and, of course, you would be using sudo with 'apt-get install') so there's no choice.

I don't know a good fix for this unfortunately.
I guess we would need to change

1) docker/mkimage
2) docker/run

which, I think, are the only occurences of sudo that execute on the host machine instead of in the docker container.

gww-parity:
Do do I need sudo for building your image?

I build many images on daily basis without sudo.

Only yours require sudo.

alastairreid:
Are you building on a Mac or a Linux machine?

I am not a docker expert but I was under the impression that you
needed sudo on a Linux machine.
If on a Mac, try removing sudo from the files I mentioned.
Let me know if that works.

--
Alastair

gww-parity:
I am building on Linux machine.
I configured my system and Docker appropriately to not need sudo.
I have both machines, one with Archlinux and another with Ubuntu,
and I successfully configured on both systems that I don't need sudo to do `docker build`.
So far, I build all other Dockerfiles so far without sudo/root, from others and my own.
Only yours required sudo/root for some reason and when I removed `sudo` from your script it didn't work out.

alastairreid:
Unfortunately, this is the first project I have ever used docker on and I
am doing so by a process of trial and error.
(This may have been obvious to you already :-))
If there is a (portable) way of using docker that avoids this problem, it
would be great to know.

alsuren:
> If there is a (portable) way of using docker that avoids this problem, it
> would be great to know.

I opened #140 that *should* do what you want.


-- 

In separate but related news, I was looking into how `cross` manages to distribute pre-built images on dockerhub with only the root user, but avoids the problem of creating files with the wrong permissions on the host filesystem. After a bit of digging, I worked out that it does something like:

[...]

PR #140, opened 2021-05-25
only sudo docker if not in docker group

alsuren:
From the discussion on #139 and reading around, I suspect that it will be fine to skip using sudo if the user is able to write to the docker socket.

On linux, https://docs.docker.com/engine/install/linux-postinstall/
suggests creating a 'docker' group to avoid the need to sudo.
Most linux distributions do this for you, so all you need to
do is ensure that your user is in the docker group for passwordless
docker usage.

I also changed `$*` to `"$@"` and a few other changes that `shellcheck` proposed on lines that I was changing. I can do a full shellcheck sweep as a separate PR if you would like, following https://google.github.io/styleguide/shellguide.html

### DIFF
--- a/docker/mkimage
+++ b/docker/mkimage
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+source $(dirname "$(realpath -s "$0")")/sudo_if_needed.bash
+
 readonly DOCKER_IMAGE_NAME="$1"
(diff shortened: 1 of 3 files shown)


alastairreid (approved):
Thank you very much for this: both the script and the fact that I can avoid this pain by adding myself to the docker group.

PR #142, opened 2021-06-04
cargo-verify: add --bin option

alastairreid:
This option allows us to build specific binaries in crates that contain [[bin]] sections.


fshaked (approved):
LGTM

PR #143, opened 2021-06-04
Generate keys, and store data in array.

fshaked:


alastairreid (approved):
I'm ok with this as it is - but it might be slightly better to put in a check that __key < 1000?

PR #145, opened 2021-06-28
Fix register_dtor issue

alastairreid:
Due to KLEE not supporting weak symbols, register_dtor (std/src/sys/unix/thread_local_dtor.rs) fails trying to invoke a symbolic function pointer instead of detecting that it is null and not invoking it.

This fixes the problem by defining the necessary symbol `__cxa_thread_atexit_impl`.
(It does not attempt to implement the full semantics of __cxa_atexit: we are mostly interested in bugs that occur before cleanup starts.)


fshaked (approved):
LGTM

PR #149, opened 2021-08-06
Add Github actions-based workflow to build and push Docker images

thierrymarianne:
Hello,

This is a proposal to have the Docker images built regularly 
(or on push to git branches following a specific pattern as needed)
so that they can be pulled from a registry by anyone 
who would like to run the rust verification tools 
(provided they'd be better off downloading images from a registry
instead of having to build them, which could be argued against as the resulting images are pretty heavy:
[~1.7 GB after compression](https://hub.docker.com/layers/161570518/thierrymarianne/contrib-rvt_r2ct-llvm-10/latest/images/sha256-4653e38083b81566e2c699c2b01b390f8bef741865b7de9284daa8bd4ced2468?context=repo) according to Docker registry).

More details about this pull-request can be found in this [issue](https://github.com/project-oak/rust-verification-tools/issues/150)

# Pre-merge requirements

 - [x] Remove user_uid, user_gid, username and ubuntu version params from matrix definition 
 - [ ] Replacing the target docker image names in the GitHub action configuration file :
   - for [rust base image](https://github.com/project-oak/rust-verification-tools/blob/290f2dc3ceb62399ca4036f730d529f419a876b3/.github/workflows/build-push-docker-images.yaml#L47-L50)
   - for [all other images](https://github.com/project-oak/rust-verification-tools/blob/290f2dc3ceb62399ca4036f730d529f419a876b3/.github/workflows/build-push-docker-images.yaml#L62-L63)
  - [ ] Having the maintainers declared `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` values in the project settings
  Docker hub tokens can be generated from https://hub.docker.com/settings/security

### DIFF
--- a/.github/workflows/build-push-docker-images.yaml
+++ b/.github/workflows/build-push-docker-images.yaml
@@ -12,14 +12,6 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        include:
-          - username: rvt
-            user_uid: 1000
-            user_gid: 1000
-            ubuntu_version: 20.04
-
     steps:
 
       - name: Checkout
@@ -40,10 +32,10 @@ jobs:
           context: docker/base
           push: true
           build-args: |
-            USERNAME=${{ matrix.username }}
-            USER_UID=${{ matrix.user_uid }}
-            USER_GID=${{ matrix.user_gid }}
-            UBUNTU_VERSION=${{ matrix.ubuntu_version }}
+            USERNAME=rvt
+            USER_UID=1000
+            USER_GID=1000
+            UBUNTU_VERSION=20.04
           cache-from: type=registry,ref=thierrymarianne/contrib-rvt_base:latest
           cache-to: type=inline
           tags: |
@@ -61,10 +53,6 @@ jobs:
       matrix:
         registry_namespace: ["thierrymarianne/"]
         image_prefix: [contrib-]
-        username: [rvt]
-        user_uid: [1000]
-        user_gid: [1000]
-        ubuntu_version: [20.04]
         google_test_version: [1.7.0]
         minisat_version: [37158a35c62d448b3feccfa83006266e12e5acb7]
         stp_version: [2.3.3]
@@ -105,7 +93,7 @@ jobs:
           push: true
           build-args: |
             FROM_IMAGE_FOR_SOLVERS=${{ matrix.registry_namespace }}${{ matrix.image_prefix }}rvt_base:latest
-            USERNAME=${{ matrix.username }}
+            USERNAME=rvt
             MINISAT_VERSION=${{ matrix.minisat_version }}
             STP_VERSION=${{ matrix.stp_version }}
             YICES_VERSION=${{ matrix.yices_version }}
@@ -122,7 +110,7 @@ jobs:
           push: true
           build-args: |
             FROM_IMAGE_FOR_KLEE=${{ matrix.registry_namespace }}${{ matrix.image_prefix }}rvt_solvers-llvm-${{ matrix.llvm }}:latest
-            USERNAME=${{ matrix.username }}
+            USERNAME=rvt
             GTEST_VERSION=${{ matrix.google_test_version }}
             LLVM_VERSION=${{ matrix.llvm }}
             KLEE_VERSION=${{ matrix.klee_version }}
@@ -139,7 +127,7 @@ jobs:
           push: true
           build-args: |
             FROM_IMAGE_FOR_SEAHORN=${{ matrix.registry_namespace }}${{ matrix.image_prefix }}rvt_klee-llvm-${{ matrix.llvm }}:latest
-            USERNAME=${{ matrix.username }}
+            USERNAME=rvt
             LLVM_VERSION=${{ matrix.llvm }}
             KLEE_VERSION=${{ matrix.klee_version }}
             SEAHORN_VERIFY_C_COMMON_VERSION=${{ matrix.seahorn_verify_c_common_version }}
@@ -158,9 +146,9 @@ jobs:
           push: true
           build-args: |
             FROM_IMAGE_FOR_RVT=${{ matrix.registry_namespace }}${{ matrix.image_prefix }}rvt_seahorn-llvm-${{ matrix.llvm }}:latest
-            USERNAME=${{ matrix.username }}
-            USER_UID=${{ matrix.user_uid }}
-            USER_GID=${{ matrix.user_gid }}
+            USERNAME=rvt
+            USER_UID=1000
+            USER_GID=1000
             RUSTC_VERSION=${{ matrix.rustc_version }}
           cache-from: type=registry,ref=${{ matrix.registry_namespace }}${{ matrix.image_prefix }}rvt-llvm-${{ matrix.llvm }}:latest
           cache-to: type=inline
@@ -174,9 +162,9 @@ jobs:
           push: true
           build-args: |
             FROM_IMAGE_FOR_RVT_R2CT=${{ matrix.registry_namespace }}${{ matrix.image_prefix }}rvt-llvm-${{ matrix.llvm }}:latest
-            USERNAME=${{ matrix.username }}
-            USER_UID=${{ matrix.user_uid }}
-            USER_GID=${{ matrix.user_gid }}
+            USERNAME=rvt
+            USER_UID=1000
+            USER_GID=1000
           cache-from: type=registry,ref=${{ matrix.registry_namespace }}${{ matrix.image_prefix }}rvt_r2ct-llvm-${{ matrix.llvm }}:latest
           cache-to: type=inline
           tags: |
--- a/.github/workflows/build-push-docker-images.yaml
+++ b/.github/workflows/build-push-docker-images.yaml
@@ -1,0 +1,183 @@
+name: Build-Push-Docker-Image
+
+on:
+  push:
+    branches:
+      - build-push-docker-images
+
+jobs:
+  build-push-base-rvt-image:
+
+    name: Build and push RVT base Docker image
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        include:
+          - username: rvt
+            user_uid: 1000
+            user_gid: 1000
+            ubuntu_version: 20.04
+
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push rvt_base
+        uses: docker/build-push-action@v2.2.2
+        with:
+          context: docker/base
+          push: true
+          build-args: |
+            USERNAME=${{ matrix.username }}
+            USER_UID=${{ matrix.user_uid }}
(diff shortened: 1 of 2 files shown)


docker/mkimage
```diff
@@ -33,6 +33,11 @@ sudo_if_needed docker build \
...
   --cache-from="${DOCKER_IMAGE_NAME}:${DOCKER_VERSION}" \
   --tag="${DOCKER_IMAGE_NAME}:${DOCKER_VERSION}" \
+  --build-arg=FROM_IMAGE_FOR_SOLVERS=rvt_base:latest \
+  --build-arg=FROM_IMAGE_FOR_KLEE=rvt_solvers:latest \
+  --build-arg=FROM_IMAGE_FOR_SEAHORN=rvt_klee:latest \
+  --build-arg=FROM_IMAGE_FOR_RVT=rvt_seahorn:latest \
+  --build-arg=FROM_IMAGE_FOR_RVT_R2CT=rvt:latest \
   --build-arg=USERNAME="$DOCKER_USER" \
   --build-arg=USER_UID="$DOCKER_UID" \
   --build-arg=USER_GID="$DOCKER_GID" \
```
alsuren: This is already quite a big PR, and I think that building images as part of CI is a more important first step than fiddling with permissions, so maybe this is probably something to put on the backlog rather than fixing right now.

I checked into how `cross` does things a while back, and wrote this:

https://github.com/project-oak/rust-verification-tools/issues/139#issuecomment-847872988

(I actually came to this repo looking for that comment for an unrelated project, and stumbled on this PR. Nice work)

.github/workflows/build-push-docker-images.yaml
```diff
@@ -0,0 +1,183 @@
...
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        include:
+          - username: rvt
+            user_uid: 1000
+            user_gid: 1000
+            ubuntu_version: 20.04
```
alsuren: (optional) I'm not sure it's worth using a matrix for defining the user ids.

I don't think that this matrix will ever have more than one entry in it: if someone wants a different user id, we're probably not going to burn compute resources building images for them. The way out of that problem would be to look into the cross thing (see my other comment).

.github/workflows/build-push-docker-images.yaml
```diff
@@ -0,0 +1,183 @@
...
+          push: true
+          build-args: |
+            USERNAME=${{ matrix.username }}
+            USER_UID=${{ matrix.user_uid }}
+            USER_GID=${{ matrix.user_gid }}
+            UBUNTU_VERSION=${{ matrix.ubuntu_version }}
+          cache-from: type=registry,ref=thierrymarianne/contrib-rvt_base:latest
+          cache-to: type=inline
+          tags: |
+            thierrymarianne/contrib-rvt_base:latest
```
alsuren, line 42: reminder to change this to change this to a repo that is controlled by the rvt team (DOCKERHUB_USERNAME and DOCKERHUB_TOKEN will need to be provided as github secrets in order for this to build once merged upstream)

alsuren (approved):
Docker and github-actions changes look fine. 
Some of the `docker/run` changes look unrelated, but otherwise seem fine.

.github/workflows/build-push-docker-images.yaml
```diff
@@ -0,0 +1,183 @@
+name: Build-Push-Docker-Image
+
+on:
+  push:
+    branches:
+      - build-push-docker-images
```
alsuren, line 6: reminder to make this trigger on tag or push-to-master or something (seems to be being used to trigger workflow runs over at https://github.com/thierrymarianne/contrib-rust-verification-tools/actions )

docker/mkimage
thierrymarianne: Indeed, I'm sorry about the verbosity. 

I can split this PR into smaller new ones easier to deal with if you'd like. 

If I recall correctly, the permissions-related environment variables were exposed for overloading in [docker/mkimage](https://github.com/project-oak/rust-verification-tools/blob/main/docker/mkimage#L16-L18), therefore I've tried to preserve them in the build script to prevent having root user and 0 as user_uid by defaut.

.github/workflows/build-push-docker-images.yaml
thierrymarianne: Indeed, the image build remaining quite resource-demanding, I didn't know which event would make more sense to you folks.

.github/workflows/build-push-docker-images.yaml
thierrymarianne: Sure thing, I'll have them written below.

docker/mkimage
alsuren: > Indeed, I'm sorry about the verbosity.

This PR is about as big as I was expecting this pull request to be (which is part of the reason why I decided not to try writing it ;-D )

I think you've made the right choice about how much to include in this PR. Hopefully the maintainer agrees, and merges it next week.

thierrymarianne:
> Docker and github-actions changes look fine.
> Some of the `docker/run` changes look unrelated, but otherwise seem fine.

This [section](https://github.com/project-oak/rust-verification-tools/blob/290f2dc3ceb62399ca4036f730d529f419a876b3/docker/run#L66-L69) would enable using the current `docker/run` logic as is (from the existing documentation and articles from the blog) in combination with a docker image pulled from a registry under the umbrella of the project.

Please let me know if you'd rather have these `docker/run` changes submitted for review with another PR.

Thank you for your feedback, @alsuren!

docker/mkimage
thierrymarianne: Regarding the permissions-related build arguments, this PR didn't introduce them though 🤷🏽

.github/workflows/build-push-docker-images.yaml
```diff
@@ -0,0 +1,171 @@
...
+
+    needs: build-push-base-rvt-image
+
+    name: Build and push RVT solvers, Klee, SeaHorn, Rust2calltree images
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        registry_namespace: ["thierrymarianne/"]
```
alastairreid, line 54: Since this repo has a liberal license, it is fine for you to do this in a fork but, unfortunately, we cannot merge this line into our repo because:

1) Any namespace would have to be a Google-owned namespace instead of "thierrymarianne/"
2) Even if we had a google-owned namespace at docker hub, I am told that the right way for a google project to do this is to publish the image in the google-cloud-container-registry
3) Publishing the image in the google-cloud-container-registry is hard for us to do for reasons.

However, this looks like a useful change for anybody else to have so maybe I could leave the issue open so that people can find this change and decide whether it is right for them?

.github/workflows/build-push-docker-images.yaml
```diff
@@ -0,0 +1,183 @@
...
+
+    name: Build and push RVT base Docker image
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        include:
+          - username: rvt
+            user_uid: 1000
```
alastairreid: Does this depend on the user_uid/gid matching the user's uid/gid?
On my personal laptop this is true: I am 1000/1000.
But on my work laptop, I have different uid/gid.

My concern with using this uid/gid is that when I use the existing docker image, invoking cargo-verify or invoking KLEE directly creates files in the host filesystem with whatever uid/gid the docker image uses.
If those don't match the uid/gid in the host machine (eg when on my work laptop), the files have invalid uid/gid values and I couldn't even delete them.
Our solution was to copy the uid/gid from the host account into docker image.

I think that our current solution to the problem is wrong (it breaks the whole idea of shared images on docker hub).
But I don't understand Docker well enough to know whether this works either.

alastairreid:
Thank you very much for this - we really appreciate it.

In particular, you have probably noticed that our use of docker is... unusual.
That is because I had never used docker before this project and I didn't know how it was meant to be used :-)

I think it would be worth splitting this into several separate PRs

1. A simple commit made up of the cleanup commits

   .dockerignore, docker/mkimage
   and the change to remove the hardcoded FROM lines in docker/*/Dockerfile and docker/mkimage
   (The latter will require the commit "Amend Dockerfiles for building images with GitHub Actions" to be split)

   I think that these are all cleanups with no change in functionality so very easy to review and LGTM.

   ps Thanks for the trick to remove hardcoded FROM lines. I had been frustrated by the hardcoding but I didn't know the solution.

2. The commit "Add more options to docker/run command"

    I *think* that this is all fine.
    But, I found it a bit hard to understand the changes due to the reorganization of the code so I am not yet *sure" it is fine.
    I am happy to take another look.

3. The most useful but also most difficult commit "Add Github actions-based workflow to build and push Docker images"

    As I said in a comment, as a Google project, we need to do everything under accounts, registries, etc. that are owned by Google.
    But, I am told that the right way for a google project to do this is to publish the image in the google-cloud-container-registry
    But, publishing the image in the google-cloud-container-registry is hard for us to do for... reasons.

    I am not entirely sure how to move this forward.
    Having a docker image is clearly a really useful thing and I am very happy that you have created this.
    You are certainly free to push an image you create to docker hub and people are free to decide whether they want to use it.

So maybe we should aim to get the easy changes (1) and (2) merged and then (3) will, at least, be simpler.

.github/workflows/build-push-docker-images.yaml
thierrymarianne: I was counting on seeing the namespace updated and I understand this might not be trivial enough to be merged.

.github/workflows/build-push-docker-images.yaml
thierrymarianne: As the user uid and user gid are declared in this section, the built image might retain this numbers.
I believe it could be an issue in the case you've described
(with another machine, different uid / gid and a mounted directory).
However there seems to be a way how to remap user uid / gid :
https://docs.docker.com/engine/security/userns-remap/#about-remapping-and-subordinate-user-and-group-ids

I have to confess that I've never had to use remapping as I usually don't modify the files mounted in containers as a host user (and when I do I try to have a host user and container user belonging to a group having the same gid with write access for said-group).

thierrymarianne:
Thank you very much for your feedback, @alastairreid.
I can follow the plan you've suggested.

thierrymarianne:
1. is ready for review at https://github.com/project-oak/rust-verification-tools/pull/151

Given that 2. was mainly about allowing to overload the docker images names, one would like to run the tools from and 
1. introduces a way how to take care of that with additional shell variables that can be overloaded,
I believe it would not bring much value anymore to merge the `docker/run` changes. What do you think?

thierrymarianne:
I've resolved the conflict for this PR to be compatible with the current main branch state for the record.
Since 3. appears to be out of reach, it might be better to close it along with [its related issue](https://github.com/project-oak/rust-verification-tools/issues/150). 

Searching closed issues or pull-requests with the right keywords would allow interested parties to find those.
Would it be ok for you, @alastairreid?

thierrymarianne:
👋🏼 @alastairreid, @alsuren,

I'm happy to let you know that some of my collaborators and I have been prototyping a web application aiming at making easier the verification of rust-based programs. 

The backend of this prototype is currently based on the rust verifications tools and  
more details can be found in the [SafePKT](https://github.com/LedgerProject/safepkt) repository documentation.

This is [how the MVP looks like](https://github.com/LedgerProject/safepkt/blob/main/docs/00-preview.md) at the moment.

alastairreid:
Hi Thierry,

This is awesome! Thanks for letting us know about this - it's great to hear
about people using our work!

--
Alastair

thierrymarianne:
This is a very early prototype but nonetheless without RVT,
we would probably have far more reasons to be scratching our heads 😅 so 🙏🏼 🙏🏼 🙏🏼 🙏🏼 
🕺🏼 🎉 💃🏼

PR #153, opened 2021-09-09
Add rudimentary support for the SMACK verifier

keram88:
Hello,

This adds rudimentary support for the SMACK verifier (https://github.com/smackers/smack) for use with the `verification-annotations` package. This includes installation of the tool and support for the verification annotations.

Currently, SMACK can run on the `t0`, `t1` and `t2` tests, but needs more work to not crash on the others.


alastairreid (approved):
Thank you very much for this!
This looks great.

